PR #69, opened 2021-03-29
API-6015-metadata-controller

jhulbert-va:
# [API-6015](https://vajira.max.gov/browse/API-6015)


bschofield-va:
IT to make sure that metadata is available?

eclendenning-va (dismissed):

vista-fhir-query-tests/src/test/java/gov/va/api/health/vistafhirquery/tests/r4/ConformanceIT.java
```diff
@@ -0,0 +1,22 @@
...
+import gov.va.api.health.r4.api.information.WellKnown;
+import gov.va.api.health.r4.api.resources.CapabilityStatement;
+import gov.va.api.health.vistafhirquery.tests.TestClients;
+import org.junit.jupiter.api.Test;
+
+public class ConformanceIT {
+
+  @Test
+  void metadata() {
+    TestClients.r4()
```
bschofield-va, line 12: This is using an test client that provides authentication headers. If well known or metadata is erroneously protected, this test will pass. It should use a test client that does not pass the static access token.

bschofield-va: `TestClients.unauthenticated()` or something.

eclendenning-va (approved):

PR #70, opened 2021-03-30
API 5589 packed ids

bschofield-va:
- wip - binary encoder, but it sux:
- wip - working smaller ids for labs
- Shorter IDs for Labs (possible through dynamically determined variable ID formats)


vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/observation/ObservationLabSamples.java
```diff
@@ -19,69 +19,6 @@
```
bschofield-va, line 19: This file looks like changed significantly, but I only added `s` to the IDs to match the update ID packing. The of the changes are from arranging the file.

jhulbert-va (approved):
LGTM

jhulbert-va (approved):

PR #74, opened 2021-04-06
Add handling for 401, 403, 408, and 500 status from charon

ilaflamme-va:


vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/WebExceptionHandlerTest.java
```diff
@@ -53,6 +57,56 @@ void badRequest() {
...
 
+  @Test
+  void forbidden() {
+    HttpClientErrorException forbidden =
+        HttpClientErrorException.Forbidden.create(HttpStatus.FORBIDDEN, null, null, null, null);
+    OperationOutcome outcome =
+        new WebExceptionHandler("").handleForbidden(forbidden, mock(HttpServletRequest.class));
+    assertThat(outcome.id(null).extension(null))
+        .isEqualTo(
+            OperationOutcome.builder()
```
bschofield-va, line 68: These OOs seem to differ in _code_ only, perhaps avoid the duplication with an `operationOutcome("forbidden")` helper

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/WebExceptionHandler.java
```diff
@@ -207,6 +218,12 @@ public OperationOutcome handleSnafu(Exception e, HttpServletRequest request) {
     return responseFor("exception", e, request, emptyList(), true);
   }
 
+  @ExceptionHandler({HttpClientErrorException.Unauthorized.class})
```
jhulbert-va, line 221: I think we wanted some sort of javadoc/ comment explained why we're mapping 401 to 500 right?

jhulbert-va: I see it in forbidden, should we just group those together with the log?

ilaflamme-va: The log statement works, ya?

jhulbert-va: I mean should it move to the ISE section and use the same log statement forbidden does?

jhulbert-va (approved):

PR #78, opened 2021-04-27
Api 6607 multiple principals

ilaflamme-va:
Do NOT merge until on-start and AWS params are updated


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/config/VistaApiConfig.java
```diff
@@ -32,7 +43,56 @@
...
+  @PostConstruct
+  void loadSiteSpecificInfoFromFile() {
+    ObjectMapper mapper = new ObjectMapper();
+    JsonNode root;
+    try (FileInputStream fs = new FileInputStream(principalFile)) {
+      root = mapper.readTree(fs);
+    } catch (FileNotFoundException e) {
+      throw new IllegalArgumentException("Failed to find principals file: " + principalFile);
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Failed to parse json file: " + principalFile);
```
bschofield-va, line 70: IO exception does not necessarily mean failed to _parse_ it. For example, If permissions are goofed, it could mean you can't open it. Honestly, I'd be inclined to `@SneakThrows` and let whatever problems happen here just kill the launch.

vista-fhir-query/spotbugs-excludes.xml
```diff
@@ -13,4 +13,9 @@
       <Bug pattern="EI_EXPOSE_REP2"/>
     </Or>
   </Match>
+  <!-- To support local testing, we allow the principal-file.json file to be overwritten by user input -->
```
jhulbert-va, line 16: I think Bryan mentioned it yesterday, but I think this can be improved upon.:
```
To support local testing, we allow the vista.api.principal-file property to be overwritten by user input.
In a real-world environment, the property will be loaded once on start from application.properties preventing user input from affecting application behavior.
```

src/scripts/make-configs.sh
```diff
@@ -54,16 +54,21 @@ main() {
...
-  requiredParam VISTA_ACCESS_CODE "${VISTA_ACCESS_CODE}"
-  requiredParam VISTA_VERIFY_CODE "${VISTA_VERIFY_CODE}"
   requiredParam VFQ_DB_URL "${VFQ_DB_URL}"
   requiredParam VFQ_DB_USER "${VFQ_DB_USER}"
   requiredParam VFQ_DB_PASSWORD "${VFQ_DB_PASSWORD}"
+  requiredParam VFQ_DEFAULT_ACCESS_CODE "${VFQ_DEFAULT_ACCESS_CODE}"
+  requiredParam VFQ_DEFAULT_VERIFY_CODE "${VFQ_DEFAULT_VERIFY_CODE}"
+  requiredParam VFQ_673_ACCESS_CODE "${VFQ_673_ACCESS_CODE}"
+  requiredParam VFQ_673_VERIFY_CODE "${VFQ_673_VERIFY_CODE}"
+  requiredParam VFQ_673_APU "${VFQ_673_APU}"
```
jhulbert-va, line 64: Standard and APU vars already exist in `shanktopus-configuration` that can be used here.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/config/VistaApiConfig.java
```diff
@@ -32,7 +38,71 @@
...
+
   public String getApplicationProxyUserContext() {
     return trimToNull(applicationProxyUserContext);
   }
+
+  /**
+   * Get the RpcPrincipal for the given site. If it doesn't exist, return the default RpcPrincipal.
+   */
+  public RpcPrincipal getSiteSpecificPrincipalOrDefault(String site) {
+    if (getSiteSpecificPrincipals() == null) {
```
jhulbert-va, line 62: Instead of a null check here, why not have `siteSpecificPrincipals()` return an empty map if it's null, and this method could just `siteSpecificPrincipals().getOrDefault(site, defaultPrincipal())`? That would make both methods null safe.

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/config/VistaApiConfigTest.java
```diff
@@ -0,0 +1,63 @@
+package gov.va.api.health.vistafhirquery.service.config;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import gov.va.api.lighthouse.charon.api.RpcPrincipal;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Test;
+
+@Slf4j
+public class VistaApiConfigTest {
```
jhulbert-va, line 10: What about checks on the default principal value?

src/scripts/make-configs.sh
bschofield-va: My `secrets.conf` leverages that.

```
. $SHANKTOSECRETS/vistalink/vista-docker.conf
VISTA_API_URL=http://localhost:8050
VFQ_DB_URL='jdbc:sqlserver://localhost:1433;sendStringParametersAsUnicode=false;database=dq'
VFQ_DB_USER='XXX'
VFQ_DB_PASSWORD='XXX'
VFQ_DEFAULT_ACCESS_CODE="$VISTA_STANDARD_USER_ACCESS_CODE"
VFQ_DEFAULT_VERIFY_CODE="$VISTA_STANDARD_USER_VERIFY_CODE"
VFQ_673_ACCESS_CODE="$VISTA_APP_PROXY_ACCESS_CODE"
VFQ_673_VERIFY_CODE="$VISTA_APP_PROXY_VERIFY_CODE"
```

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/config/VistaApiConfig.java
bschofield-va: 10/10 collections should never be null return values... empty is just fine, but null is no bueno

No one want to do this: 
```
var things = getThing()
if (things == null || things.isEmpty()) { ... }
```

ilaflamme-va: I'll add a lazy getter.

jhulbert-va (approved):

eclendenning-va (approved):

bschofield-va (approved):

PR #87, opened 2021-06-08
Organization searchRaw endpoint

ilaflamme-va:
~dependent on the [basic-rpc-model](https://github.com/department-of-veterans-affairs/lighthouse-charon/pull/90)~

search raw endpoint for getting back xml from an rpc.
requires site and icn.


jhulbert-va (dismissed):
Protect this boi with a client-key

ilaflamme-va:
Client-key protect is now in place. Created new controller to house `/raw/{Resource}`

bschofield-va:
Some sort of tests to make sure this endpoint is available and requires a client key please

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/raw/RawController.java
```diff
@@ -0,0 +1,34 @@
...
+    value = "/raw",
+    produces = {"application/json", "application/fhir+json"})
+@AllArgsConstructor(onConstructor_ = {@Autowired, @NonNull})
+@Builder
+public class RawController {
+
+  private final VistalinkApiClient vistalinkApiClient;
+
+  @GetMapping(
+      value = "/Organization",
```
jhulbert-va, line 28: This should include the fhir version I think.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/config/ClientKeyProtectedEndpointConfig.java
```diff
@@ -50,7 +49,7 @@
             .unauthorizedResponse(unauthorizedResponse())
             .build());
 
-    registration.addUrlPatterns("/internal/*", PathRewriteConfig.leadingPath() + "internal/*");
+    registration.addUrlPatterns("/raw/*", PathRewriteConfig.leadingPath() + "raw/*");
```
jhulbert-va, line 52: I believe in the past we prefixed any internal path with `/internal`. I think my vote would be to leave this and prepend internal to the path in the controller.

bschofield-va: Are you preposing `/internal/raw/Organization`?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/raw/RawController.java
jhulbert-va: What happens if we ever support more than one version like DQ does?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/config/ClientKeyProtectedEndpointConfig.java
jhulbert-va: Yes.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/raw/RawController.java
eclendenning-va: then it will happen after the transformation. Even in dq we make r4 and dstu2 payloads from the same  _raw_ data

jhulbert-va: 🤦 😬

bschofield-va (approved):

PR #96, opened 2021-06-23
Api 8090 use new rpc principals

eclendenning-va:


src/scripts/make-configs.sh
```diff
@@ -194,6 +199,31 @@ populatePrincipalFile() {
...
 }
 
+populateRpcPrincipalFile() {
+  cat > $REPO/vista-fhir-query/config/principals-$PROFILE.json << EOF
+{
+    "entries" : [
+        {
+            "rpcNames" : [
+                "VPR GET PATIENT DATA",
+                "IBLHS AMCMS GET INS"
```
jhulbert-va, line 209: :lyin:
IBLHS has the same A/V codes, but a different APU.

vista-fhir-query/config/.gitignore
```diff
@@ -1,2 +1,3 @@
 *.properties*
-principal.json
\ No newline at end of file
+principal.json
```
jhulbert-va, line 2: Do we just want to ignore all files here?

bschofield-va:
This also needs a DU update, right?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/RestVistaApiClient.java
```diff
@@ -54,8 +58,9 @@ public RpcResponse requestForTarget(
       RpcVistaTargets target, TypeSafeRpcRequest rpcRequestDetails) {
     RpcRequest rpcRequest =
         RpcRequest.builder()
-            .principal(config().defaultPrincipal())
-            .siteSpecificPrincipals(config().siteSpecificPrincipals())
+            .principal(RpcPrincipal.builder().accessCode("not-used").verifyCode("not-used").build())
+            .siteSpecificPrincipals(
+                rpcPrincipalLookup.findByName(rpcRequestDetails.asDetails().name()))
```
bschofield-va, line 63: If this request is for a specific site, then we will be sending over _all_ of the credentials, despite not needing them. It's not great, but I'm ok to move forward as is because of the site-specific API changes that I think are coming.

bschofield-va (approved):

PR #97, opened 2021-06-25
Api 8090 loma linda hack

eclendenning-va:


src/scripts/make-configs.sh
```diff
@@ -63,6 +63,8 @@ main() {
   requiredParam VFQ_673_VERIFY_CODE "${VFQ_673_VERIFY_CODE}"
   requiredParam VFQ_673_APU "${VFQ_673_APU}"
   requiredParam VISTA_API_VPR_GET_PATIENT_DATA_CONTEXT "${VISTA_API_VPR_GET_PATIENT_DATA_CONTEXT}"
+  requiredParam VISTA_API_LOMA_LINDA_HACK_CONTEXT "${VISTA_API_LOMA_LINDA_HACK_CONTEXT}"
```
jhulbert-va, line 66: Locally this wouldn't be needed.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/RestVistaApiClient.java
```diff
@@ -43,6 +44,17 @@
         .body(body);
   }
 
+  private Map<String, RpcPrincipal> getPrincipals(TypeSafeRpcRequest rpcRequestDetails) {
+    Map<String, RpcPrincipal> principals =
+        rpcPrincipalLookup.findByName(rpcRequestDetails.asDetails().name());
+    // Loma Linda context hack
```
jhulbert-va, line 50: Log statement instead so we don't lose track of it? "Performing Loma-Linda context override."

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/RestVistaApiClient.java
```diff
@@ -43,6 +44,17 @@
...
+    RpcPrincipal maybeLomaLinda = principals.get("605");
+    if (maybeLomaLinda != null) {
+      principals.put("605", maybeLomaLinda.contextOverride(config.getLomaLindaHackContext()));
```
jhulbert-va, line 53: This could set the context to `null`... Or rather "unset", but I think it's better to try with the default than either of those.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/RestVistaApiClient.java
```diff
@@ -43,6 +44,17 @@
         .body(body);
   }
 
+  private Map<String, RpcPrincipal> getPrincipals(TypeSafeRpcRequest rpcRequestDetails) {
+    Map<String, RpcPrincipal> principals =
+        rpcPrincipalLookup.findByName(rpcRequestDetails.asDetails().name());
+    // Loma Linda context hack
+    RpcPrincipal maybeLomaLinda = principals.get("605");
+    if (maybeLomaLinda != null) {
```
ilaflamme-va, line 52: This will add a context override for all requests using Loma Linda. That will include future resources that are worked.

This is ok for now, because we only use the `VPR GET PATIENT DATA` with VFQ.

jhulbert-va (approved):

PR #102, opened 2021-07-02
API-8127-coverage-transformer

jhulbert-va:
# [API-8127](https://vajira.max.gov/browse/API-8127)


bschofield-va (approved):

bschofield-va:
I'll follow up with some minor changes in the sibling stories to prevent too much conflict

PR #113, opened 2021-07-15
API-8284-finalize-read-docs

jhulbert-va:
# [API-8284](https://vajira.max.gov/browse/API-8284)
This may be as finalized as we get this for now, but I figure we can use this draft PR for discussion and updates if we think we need them.


jhulbert-va:
Will re-open once finalizing of write is also complete.

PR #131, opened 2021-08-06
API-000: Fix connect timeout for resttemplate

jabando-va:


bschofield-va (changes requested):
yo homie, where's your test.
fix a bug... write a test to prove a) you can reproduce it, and b) you fixed it.

bschofield-va (approved):

PR #133, opened 2021-08-16
API-8865: coverageclass url

jabando-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageWitnessProtectionAgent.java
```diff
@@ -29,8 +29,8 @@
                     c ->
                         ProtectedReference.builder()
                             .type("InsurancePlan")
-                            .id(c.value())
-                            .onUpdate(c::value)
+                            .id(c.value().substring("InsurancePlan/".length()))
```
jhulbert-va, line 32: If we just have to replace it here anyway, I almost wonder if adding `InsurancePlan/` just gets done here and not in the transformer. 

@bschofield-va thoughts?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageTransformer.java
```diff
@@ -62,7 +62,7 @@ static boolean stopPolicyFromBillingToBoolean(String value) {
         .map(
             coords ->
                 Coverage.CoverageClass.builder()
-                    .value(coords.toString())
+                    .value("InsurancePlan/" + coords)
```
jhulbert-va, line 65: The coordinates need to use `toString()` in order for them to be in the correct string format.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageWitnessProtectionAgent.java
bschofield-va: Yeah... I don't really like the `InsurancePlan/` prefix knowledge spread across two places: Easy to introduce a bug when we decide we want a different prefix. However ... if this bit were smart enough to _detect_ `Any/Prefix/Here/12345` and dynamically remove, then re-add it, we'd be in good shape. The prefix logic would be in the transformer, where it belongs and this thing just focuses on updating the ID number.

bschofield-va: I think the logic is
- look for  a `/` 
  - if not found, just witness protect the whole value
  - if found,
    - separate prefix and ID based on last occurrence of `/`
    - witness protect the ID
    - save by re-appending the prefix

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageWitnessProtectionAgent.java
```diff
@@ -29,8 +53,8 @@
                     c ->
                         ProtectedReference.builder()
```
jhulbert-va, line 54: Instead of splitting it twice in two different functions, why not just do it here?
```
c -> {
  splitToPrefixAndCode
  // Using prefix and code
  return ProtectedReference
}
```

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageWitnessProtectionAgent.java
```diff
@@ -17,6 +19,28 @@
...
+    if (!id.contains("/")) {
+      return id;
+    }
+    List<String> strings = Arrays.asList(id.split("/"));
+    return strings.get(strings.size() - 1);
+  }
+
+  /** Get prefix of a string, if it has one. */
+  public String getPrefix(String id) {
+    if (!id.contains("/")) {
```
jhulbert-va, line 33: If I don't find a `/`, I for sure don't want to just clobber the whole value, but I also don't want to leave it as is. Instead, why not just encode the whole value?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageWitnessProtectionAgent.java
```diff
@@ -17,6 +19,28 @@
...
+      return "";
+    }
+    List<String> strings = Arrays.asList(id.split("/"));
+    StringBuilder prefix = new StringBuilder("/");
```
jhulbert-va, line 37: We don't want to be in the business of splitting a string just to reappend it all using a string builder.

jhulbert-va (approved):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageWitnessProtectionAgent.java
```diff
@@ -50,12 +26,14 @@ public String getPrefix(String id) {
...
-                            .type("InsurancePlan")
-                            .id(getId(c.value()))
-                            .onUpdate(i -> c.value(getPrefix(c.value()) + i))
-                            .build()));
+                    c -> {
+                      int indexOfId = c.value().lastIndexOf('/') + 1;
+                      return ProtectedReference.builder()
+                          .type("InsurancePlan")
+                          .id(c.value().substring(indexOfId))
+                          .onUpdate(i -> c.value(c.value().substring(0, indexOfId) + i))
```
jhulbert-va, line 34: You're gonna end up throwing here if the id doesn't contain a `/`. `lastIndexOf` would return `-1` plus 1 is 0, so you'd end up with `substring(0, -1)`. You'll want to make this safer? Perhaps create a basic builder with your static value(s) and then build the rest based on the index value?

jhulbert-va: Speaking of which, make sure there's a test for this.

jhulbert-va (approved):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageWitnessProtectionAgent.java
```diff
@@ -26,12 +26,14 @@
...
             resource.coverageClass().stream()
                 .map(
-                    c ->
-                        ProtectedReference.builder()
-                            .type("InsurancePlan")
-                            .id(c.value())
-                            .onUpdate(c::value)
-                            .build()));
+                    c -> {
+                      int indexOfId = c.value().lastIndexOf('/') + 1;
```
bschofield-va, line 30: danger! _never_ access an index without verifying it exists.
`/this/will/explode/InsurancePlan/1234/`

bschofield-va (approved):

PR #148, opened 2021-09-09
API-9862-post-stub

jhulbert-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/api/R4CoverageApi.java
```diff
@@ -7,19 +7,74 @@
...
+        responseCode = "403",
+        description = "Forbidden",
+        content = {
+          @Content(
+              mediaType = "application/fhir+json",
+              schema = @Schema(implementation = OperationOutcome.class))
+        })
+  })
+  ResponseEntity<String> coverageCreate(
+      @Parameter(
```
bschofield-va, line 54: We don't need the HttpServletRequest?

vista-fhir-query/pom.xml
```diff
@@ -15,11 +15,15 @@
     <fhir-resources.version>8.0.11</fhir-resources.version>
     <jacoco.coverage>0.94</jacoco.coverage>
     <mssql-jdbc.version>9.2.1.jre15</mssql-jdbc.version>
-    <swagger-maven-plugin.version>2.1.9</swagger-maven-plugin.version>
+    <swagger-maven-plugin.version>2.1.10</swagger-maven-plugin.version>
     <talos.version>1.0.0</talos.version>
     <swagger-examples-maven-plugin.version>3.0.0</swagger-examples-maven-plugin.version>
   </properties>
   <dependencies>
+    <dependency>
```
bschofield-va, line 23: why this?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/api/R4CoverageApi.java
jhulbert-va: Not sure what we'd need it for at this point. We should be able to build the charon request with the values here.

bschofield-va (approved):

PR #150, opened 2021-09-14
API-9899-coverage-write-transformer

jhulbert-va:
Two things on this PR:
1. Going from the "parent" relationship code in fhir to a vista code becomes ambiguous (thread [here](https://libertyits.slack.com/archives/GBULYN18X/p1631552805375800)). However, I figured the better way to deal with it would be to push this through and come back around later once we have an answer on how we will deal with that ambiguity.
2. ~I was not 100% sure on the `safelyAdd()` function. It follows the normal convention of fhir (give what we can no matter what), but it may be safer for us to throw when we know something is wrong vs waiting on vista to kick it back.~

Thoughts?


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/R4Transformers.java
```diff
@@ -59,6 +59,19 @@ public static CodeableConcept asCodeableConcept(Coding coding) {
     return filtered.isEmpty() ? null : filtered;
   }
 
+  /** Given a reference, get the id and try to parse it as record coordinates. */
+  public static Optional<RecordCoordinates> getRecordCoordinates(IsReference maybeReference) {
```
bschofield-va, line 63: I really dislike `get` as a method prefix given the heavy weight that `getX`, `setX`, and `isX` carry with the JavaBean convention. `get` and `set` are pretty tainted.  I see that we let `getReferenceId` slide in this class, but let's not add to the badness by propagating the same sin on new stuff.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/RelationshipMapping.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.common.collect.ImmutableMultimap;
+import gov.va.api.health.r4.api.datatypes.Coding;
+import gov.va.api.lighthouse.charon.models.lhslighthouserpcgateway.LhsLighthouseRpcGatewayResponse.UnexpectedVistaValue;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NonNull;
+
+public class RelationshipMapping {
+
+  private static final ImmutableMultimap<String, Relationship> RELATIONSHIPS =
```
bschofield-va, line 12: why is this a multimap, it's populated 1-1.

bschofield-va: Also, the name could be better... what is the string that is the map key? File? Field? who knows?!

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageTransformer.java
```diff
@@ -108,6 +117,22 @@ private CodeableConcept coverageClass() {
     return extensions.isEmpty() ? null : extensions;
   }
 
+  private Identifier memberId(String memberId) {
```
bschofield-va, line 120: why this?

bschofield-va:
Not so sure the `safelyAdd` is good. If we can validate _anything_ before we send it to Vista we should and kick back a `400` response. I _need_ VistA to be our safety net, but that doesn't mean we should willingly and knowingly throw garbage at it.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageTransformer.java
jhulbert-va: While going through Jays spreadsheet to populate the write transformer, I noticed we weren't mapping it in the read.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/RelationshipMapping.java
jhulbert-va: They're ever so slightly not 1-to-1.

I fix the naming problem.

jhulbert-va: I'll make this an enum and base it on the vista values (e.g. there would be a mother and father enum value that both map to a `parent` fhir code.

jhulbert-va: Found stuff we were mapping incorrectly as well, so I'll need to add values for those as well.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageTransformer.java
bschofield-va: S C O P E C R E E P

Please don't muddly a complicated PR with stuff not related to it.

bschofield-va (approved):

PR #157, opened 2021-09-17
API-9898-charon-create

jhulbert-va:
# [API-9898](https://vajira.max.gov/browse/API-9898)


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/WebExceptionHandler.java
```diff
@@ -236,6 +236,12 @@ OperationOutcome handleUnauthorized(Exception e, HttpServletRequest request) {
     return responseFor("unauthorized", e, request, emptyList(), true);
   }
 
+  @ExceptionHandler({ResourceExceptions.BadPayload.class})
```
bschofield-va, line 239: I think _Bad Payload_ is too generic. This could be mistaken for a bad _response_ payload from Charon. Perhaps _BadRequestPayload_ is more clear?

bschofield-va (approved):

PR #158, opened 2021-09-20
endpoint read

ilaflamme-va:


jhulbert-va (dismissed):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/endpoint/R4EndpointController.java
```diff
@@ -58,6 +73,10 @@
     return toBundle(request).apply(stations);
   }
 
+  private boolean isKnownSite(String site) {
+    return rpcPrincipalLookup.findByNameAndSite("LHS LIGHTHOUSE RPC GATEWAY", site).isPresent();
```
bschofield-va, line 77: Not sure how I feel about this ... thinking

bschofield-va: The hard coded string references need to be replaced with `gov.va.api.lighthouse.charon.models.lhslighthouserpcgateway.LhsLighthouseRpcGatewayGetsManifest.RPC_NAME`

and we need to think about and acknowledge that this is a little bit of a lie. This is only telling sites that support RPC Gateway, not all known sites to VFQ.

bschofield-va: i 🇷🇺

bschofield-va (approved):

PR #171, opened 2021-09-27
API-10156: CREATE InsurancePlan minimum transformer

jabando-va:


jhulbert-va (commented):
I see some repeatition between this and the Coverage transformer. See if you can move some of the repeated lines to a common place.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,228 @@
...
+        .index(index)
+        .value(value)
+        .build();
+  }
+
+  WriteableFilemanValue planStandardFtf(Extension extension, String field, String value) {
+    if (isBlank(extension) || isBlank(extension.valueQuantity())) {
+      throw ResourceExceptions.BadRequestPayload.because(field, "extension is null");
+    }
+    return pointer(field, 1, value);
```
jhulbert-va, line 99: This method feels odd. I'd expect it to know what file the pointer is to already without it being passed in.
In addition, the extension isn't being used at all except for a null check?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,228 @@
...
+        .ifPresentOrElse(
+            fields::add,
+            () -> {
+              throw ResourceExceptions.BadRequestPayload.because(
+                  GroupInsurancePlan.PLAN_STANDARD_FTF, "extension not found");
+            });
+    return fields;
+  }
+
+  WriteableFilemanValue typeOfPlan(List<InsurancePlan.Plan> plan) {
```
jhulbert-va, line 202: How many types of plans are there in this resource? If there's a bunch, we should think about pulling out some of the repeated code.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,228 @@
...
+            e ->
+                this.planStandardFtf(
+                    e,
+                    GroupInsurancePlan.PLAN_STANDARD_FTF_VALUE,
+                    e.valueQuantity().value().toString()))
+        .ifPresentOrElse(
+            fields::add,
+            () -> {
+              throw ResourceExceptions.BadRequestPayload.because(
+                  GroupInsurancePlan.PLAN_STANDARD_FTF, "extension not found");
```
jhulbert-va, line 197: This line is repeated _a lot_.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
jabando-va: `Select the type of plan that best describes this plan.  The type of policy will be used to determine if reimbursement for claims from the insurance carrier is appropriate.  It will also be used to determine what other fields and displays are appropriate for this plan.  If unknown or unsure, pick the more general type of plan.
`
Sounds like there could be a lot of types? Not sure if they are listed anywhere in the spreadsheet

jhulbert-va: They should be in the spreadsheet. I saw this and the electronic type.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
bschofield-va: This whole pattern is repeated _a lot_. This is a 100 line method... we'll need some refactoring to reduce duplication here.

bschofield-va: We also need to work on these errors. Put yourself in future JJ's place. Are you helping future JJ figure out what went wrong with these exception messages?

Customer: Why this no save?
JJ: extension not found.
Customer: Which?
JJ: \*panic*

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,228 @@
...
+              if (c.type() == null) {
+                return false;
+              }
+              return c.type().coding().stream()
+                  .anyMatch(
+                      coding ->
+                          "urn:oid:2.16.840.1.113883.3.8901.3.1.355803.8009"
+                              .equals(coding.system()));
+            })
+        .map(c -> c.type().coding().get(0).code())
```
bschofield-va, line 219: _never_ index a list or array with being 100% certain it's there.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,228 @@
...
+@Value
+public class R4InsurancePlanToInsuranceTypeFileTransformer {
+  @NonNull InsurancePlan insurancePlan;
+
+  @Builder
+  R4InsurancePlanToInsuranceTypeFileTransformer(@NonNull InsurancePlan insurancePlan) {
+    this.insurancePlan = insurancePlan;
+  }
+
+  WriteableFilemanValue electronicPlanType(List<CodeableConcept> codeableConcepts) {
```
bschofield-va, line 31: this entire method is basically repeated at the bottom...

https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

Extract what is different and consolidate.

 extract what is different

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,228 @@
...
+    if (isBlank(codeableConcepts)) {
+      throw ResourceExceptions.BadRequestPayload.because(
+          GroupInsurancePlan.ELECTRONIC_PLAN_TYPE, "plan is null");
+    }
+    return codeableConcepts.stream()
+        .filter(
+            c -> {
```
bschofield-va, line 38: this nugget also makes an appearance in Coverage. Let's consolidate across resources with some common utility methods.

The overall problem is the _intention_ is lost because of the details. Would future JJ rather read this as is or perhaps something like

```
var fhirField = "type"
var filemanField = GroupInsurancePlan.ELECTRONIC_PLAN_TYPE;
var system = "urn:oid:123.456.7890";
// ^ omg, these could method parameters!

codeableConcepts.stream()
  .filter(withCodingSystem(system)
  .map(toCodeFromFirstCoding())
  .map(toPointerToIen(filemanField))
  .findFirst()
  .orElseThrow(missingCodeableConcept(fhirFhid,system,filemanField))
```

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,228 @@
...
+              if (c == null) {
+                return false;
+              }
+              return c.coding().stream()
+                  .anyMatch(
+                      coding ->
+                          "urn:oid:2.16.840.1.113883.3.8901.3.1.355803.8015"
+                              .equals(coding.system()));
+            })
+        .map(c -> c.coding().get(0).code())
```
bschofield-va, line 48: I do question this assumption that it's the _first_ match. You filter seems to look at _all_ of the coding. If there is more than one, and the last one is what matters, you will pick the wrong one.

Perhaps you're better converting each CodeableConcept to multiple Codings, then filtering.

```
codeableConcepts.stream()
  .flatMap(CodableConcept::coding)
  .filter(codingWithSystem(...))
  ...
  ```

File: vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java [hunk 8]
```diff
@@ -0,0 +1,228 @@
...
+        .findFirst()
+        .map(id -> pointer(GroupInsurancePlan.ELECTRONIC_PLAN_TYPE, 1, id))
```
bschofield-va, line 50: Is this really an _id_ ... it was previously mapped to a _code_.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/R4Transformers.java
```diff
@@ -72,6 +72,14 @@ public static CodeableConcept asCodeableConcept(Coding coding) {
     return Optional.ofNullable(referenceParts[referenceParts.length - 1]);
   }
 
+  /** Checks if the given system value has a match in the given CodeableConcept's coding List. */
+  public static boolean hasCodingSystem(CodeableConcept c, String system) {
+    if (c == null) {
+      return false;
+    }
+    return c.coding().stream().anyMatch(coding -> system.equals(coding.system()));
```
jhulbert-va, line 80: What if coding is `null`?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/R4Transformers.java
```diff
@@ -72,6 +72,14 @@ public static CodeableConcept asCodeableConcept(Coding coding) {
     return Optional.ofNullable(referenceParts[referenceParts.length - 1]);
   }
 
+  /** Checks if the given system value has a match in the given CodeableConcept's coding List. */
+  public static boolean hasCodingSystem(CodeableConcept c, String system) {
+    if (c == null) {
```
jhulbert-va, line 77: `isBlank`

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,205 @@
...
+        "system type 'urn:oid:2.16.840.1.113883.3.8901.3.1.355803.8015' not found");
+  }
+
+  private Optional<Extension> extensionForSystem(String system, String field) {
+    if (isBlank(insurancePlan().extension())) {
+      throw ResourceExceptions.BadRequestPayload.because("extension is null");
+    }
+    return Optional.ofNullable(
+        insurancePlan().extension().stream()
+            .filter(e -> system.equals(e.url() != null ? e.url() : e.valueQuantity().system()))
```
jhulbert-va, line 49: These two systems are very different, one defines an extension and the other a type of unit. I'm not sure I like one method to match both.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,205 @@
...
+            .findFirst()
+            .orElseThrow(
+                () -> ResourceExceptions.BadRequestPayload.because(field, "extension not found")));
+  }
+
+  void extensionNullCheck(Extension extension, String field) {
+    if (isBlank(extension) || isBlank(extension.valueQuantity())) {
```
jhulbert-va, line 56: This isn't just an extension null check. It's very specific to a quantity extension.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,205 @@
...
+  WriteableFilemanValue extensionToBooleanWriteableFilemanValue(Extension extension, String field) {
+    if (isBlank(extension) || isBlank(extension.valueBoolean())) {
+      throw ResourceExceptions.BadRequestPayload.because(
+          field, "extension is null or has no boolean value");
+    }
+    return insuranceTypeCoordinates(field, 1, extension.valueBoolean() ? "YES" : "NO");
+  }
+
+  String extractCodeFromCoding(CodeableConcept c) {
+    return c.coding().stream().findFirst().isPresent()
```
jhulbert-va, line 71: This can be simplified so you arent getting the same stream twice:
```
c.coding().stream().findFirst().map(Coding::code).orElse("");
```

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  private Optional<Extension> extensionForSystem(String system, String field) {
+    return matchOnSystem(field, e -> system.equals(e.url()));
+  }
+
+  private Optional<Extension> extensionForUnitSystem(String system, String field) {
+    return matchOnSystem(
+        field, e -> e.valueQuantity() != null && system.equals(e.valueQuantity().system()));
+  }
```
jhulbert-va, line 54: Looking at this PR closer, the unit system one is wrong. All the things using it are checking for the `extension` url, not the quantity system. This will result in the field never being populated. 

Instead, we should make a common method in `R4Transformers` that allows us to filter an extension by its defining url.
```
Optional<Extension> extensionForSystem(List<Extension> extensions, String system);
```

Determining what the fileman value is from a valuequantity should then be its own method that would be used by extensions where the value is a quantity.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,251 @@
...
+        1,
+        extension.valueQuantity().value().toString());
+  }
+
+  private WriteableFilemanValue pointer(int index, String ien) {
+    if (isBlank(ien)) {
+      return null;
+    }
+    return WriteableFilemanValue.builder()
+        .file(InsuranceCompany.FILE_NUMBER)
```
jhulbert-va, line 171: This shouldn't assume that all pointers will point to the InsuranceCompany file. In reality, only one will.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,251 @@
...
+        GroupInsurancePlan.PLAN_STANDARD_FTF, 1, extension.valueQuantity().unit());
+  }
+
+  WriteableFilemanValue planStandardFtfValue(Extension extension) {
+    extensionHasQuantityNullCheck(extension, GroupInsurancePlan.PLAN_STANDARD_FTF_VALUE);
+    return insuranceTypeCoordinates(
+        GroupInsurancePlan.PLAN_STANDARD_FTF_VALUE,
+        1,
+        extension.valueQuantity().value().toString());
+  }
```
jhulbert-va, line 164: These will come from the same extension, instead of dealing with it separately, why not return a collection of `WriteableFilemanValues` and then use `results.addAll`

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
jabando-va: Im not really following the issue with the unit system method. For the unit system extensions, `extension.url` is not populated, instead, the URL is bound to `extension.valueQuantity().system()` as shown in InsurancePlanSamples.

So R4Transformers gets two new methods? extensionForSystem and extensionForUnitSystem? It sounds like I just move the implementation for these two over to that class.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/insuranceplan/R4InsurancePlanToInsuranceTypeFileTransformer.java
```diff
@@ -0,0 +1,220 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+@Value
+public class R4InsurancePlanToInsuranceTypeFileTransformer {
```
jhulbert-va, line 30: `GroupInsurancePlan` not `InsuranceType`

jhulbert-va (commented):
Still have to check this against requirements in the spreadsheet. It looks like there is gonna be some issues. I find it suspect that all of those extensions aren't required.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/R4Transformers.java
```diff
@@ -59,6 +69,35 @@ public static CodeableConcept asCodeableConcept(Coding coding) {
...
+      throw ResourceExceptions.BadRequestPayload.because("extension is null");
+    }
+    return Optional.ofNullable(
+        extensions.stream()
+            .filter(
+                e -> e.valueQuantity() != null && definingUrl.equals(e.valueQuantity().system()))
+            .findFirst()
+            .orElseThrow(
+                () -> ResourceExceptions.BadRequestPayload.because("extension not found")));
+  }
```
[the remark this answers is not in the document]
jabando-va: Its used in the Transformer though: it needs to process/filter on a `valueQuantity().system()` for an extension holding a unit value
```extensionForUnitSystem(
insurancePlan().extension(),
"http://va.gov/fhir/StructureDefinition/insuranceplan-planStandardFilingTimeFrame")
.map(this::planStandardFtf)
.ifPresent(fields::addAll);

File: vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/R4Transformers.java [hunk 19]
```diff
@@ -59,6 +69,22 @@ public static CodeableConcept asCodeableConcept(Coding coding) {
...
+  public static Optional<Extension> extensionForSystem(
+      List<Extension> extensions, String definingUrl) {
+    if (isBlank(extensions)) {
+      throw ResourceExceptions.BadRequestPayload.because("extension is null");
+    }
+    return Optional.ofNullable(
+        extensions.stream()
+            .filter(extension -> definingUrl.equals(extension.url()))
+            .findFirst()
+            .orElseThrow(
```
jhulbert-va, line 82: If you're going to return an optional, you shouldn't be throwing here. Try to find it, return the optional. Whatever uses this method will then determine what to do if it doesn't exist.

jhulbert-va (approved):

PR #172, opened 2021-09-28
API-10240-coverage-error-handling

jhulbert-va:
# [API-10240](https://vajira.max.gov/browse/API-10240)


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/LhsGatewayErrorHandler.java
```diff
@@ -0,0 +1,78 @@
...
+    RECORD_DOESNT_EXIST("601"),
+    INVALID_FIELD("701"),
+    UNKNOWN("000");
+
+    @Getter private final String code;
+
+    /** Determine an error type based on the error code. */
+    public static LhsGatewayError forCode(@NonNull String code) {
+      var matchingErrors =
+          Arrays.stream(LhsGatewayError.values()).filter(c -> code.equals(c.code())).toList();
```
bschofield-va, line 71: Why `toList()` here vs. `findFirst()`? It seems unnecessary to allocate a List object to to allocate _another_ stream object to allocate an Optional.

jhulbert-va: I wanted to make sure we were masking any problems. As this grows, we may accidentally add another matching code.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/LhsGatewayErrorHandler.java
```diff
@@ -0,0 +1,78 @@
...
+      throw ExpectationFailed.because("Errors 'code' value is null.");
+    }
+    LhsGatewayError error = LhsGatewayError.forCode(errorDetails.get("code"));
+    switch (error) {
+      case RECORD_DOESNT_EXIST -> throw CannotUpdateUnknownResource.because(
+          unknownRecordVistaCoordinates());
+      case INVALID_FIELD -> throw BadRequestPayload.because(
+          "Vista Description: " + errorDetails.get("text"));
+      case UNKNOWN -> throw new FatalServerError(errorDetails.toString());
+      default -> throw new IllegalStateException(
```
bschofield-va, line 33: Not sure I understand this choice.
If there is no `code`, then throw `ExpectationFailed`
If there is an unknown `code`, then throw a `IllegalState`

Both ^ represent an unexpected situation, but result in different, completely unrelated types of exceptions.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/LhsGatewayErrorHandler.java
```diff
@@ -0,0 +1,78 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NonNull;
+
+@AllArgsConstructor(staticName = "of")
+public class LhsGatewayErrorHandler {
+  LhsLighthouseRpcGatewayResponse.Results results;
+
+  private void handleError(Map<String, String> errorDetails) {
+    if (isBlank(errorDetails.get("code"))) {
```
bschofield-va, line 23: You're paying for the lookup twice.
```
var code = errorDetails.get("code")
```

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4SiteCoverageController.java
```diff
@@ -147,8 +144,20 @@ public void coverageUpdate(
     if (isBlank(body.id()) || !privateId.equals(body.id())) {
       throw CannotUpdateResourceWithMismatchedIds.because(privateId, body.id());
     }
-    updateOrCreate(
-        CoverageWriteContext.builder().site(site).body(body).mode(CoverageWriteApi.UPDATE).build());
+    var coordinates = PatientTypeCoordinates.fromString(privateId);
```
bschofield-va, line 147: I think we should normalize the level of abstraction here with some well named methods so intention is more clear, e.g.
```
var ctx = ...
validateIdsMatch(...)
validateSitesMatch(...)
validatePatientIcnsMatch(...) 
updateOrCreate(...)
```

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/LhsGatewayErrorHandler.java
jhulbert-va: The latter should realistically never happen, we set an unknown code to unknown. The default was to satisfy error prone. Instead of `ExpectationFailed` for the former though, I wonder if a `UnexpectedVistaValue` would be better to convey that vista gave us something we didn't like.

bschofield-va (approved):

PR #179, opened 2021-09-29
Remove trailing comma from IENs from id fields so that I3 match

bschofield-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/RecordCoordinates.java
```diff
@@ -16,6 +16,16 @@
...
   @NonNull String ien;
 
+  /**
+   * Create a new entry, removing the trailing comma from the ien. E.g., "123,456," becomes
+   * "123,456".
+   */
+  public RecordCoordinates(@NonNull String site, @NonNull String file, @NonNull String ien) {
+    this.site = site;
+    this.file = file;
+    this.ien = endsWithTrailingComma(ien) ? removeTrailingComma(ien) : ien;
```
jhulbert-va, line 26: Why not have `removeTrailingComma` do the check?

jhulbert-va (commented):
Do we need this for the PatientCoordinates as well?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/RecordCoordinates.java
bschofield-va: so my intentions are clear when setting this value

PR #181, opened 2021-09-30
API-10161-minimum-viable-transformer

jhulbert-va:
# [API-10161](https://vajira.max.gov/browse/API-10161)


bschofield-va (commented):
taking as is, any tweaks will follow up in subsequent prs

bschofield-va (approved):

PR #248, opened 2021-11-02
API-000-add-guava

jhulbert-va:


vista-fhir-query/pom.xml
```diff
@@ -35,6 +35,11 @@
       <artifactId>charon-models</artifactId>
       <version>${charon.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>31.0.1-jre</version>
```
jblefkove-va, line 41: property?

jblefkove-va (approved):
:shipit:

PR #252, opened 2021-11-03
API-000-extension-processors-support-index

jhulbert-va:

- In order to support lower level extension processing in CER, the processors need to allow an index be given.
- This index is optional and defaults to `1` when one is not given to keep backwards compatibility.


eclendenning-va (dismissed):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/extensionprocessing/AbstractExtensionHandler.java
```diff
@@ -1,16 +1,26 @@
...
 
   @Getter private final WriteableFilemanValueFactory filemanFactory;
+
+  @Getter private final int index;
+
+  protected AbstractExtensionHandler(
+      String definingUrl,
+      Required required,
+      WriteableFilemanValueFactory filemanFactory,
+      Integer index) {
```
bschofield-va, line 20: These nullable _integers_ smell ... do they really need to be specified as null?

jhulbert-va: I wrote it that way for two reasons, 1 it's a common place to create the default value (for backwards compatibility), and 2 because for resources like coverage or insurance plan, where we're creating a flat resource, we don't necessarily need to be tracking indexes past 1. Perhaps as part of the clean-up task we can update this to disallow null values?

bschofield-va: > 1 it's a common place to create the default value (for backwards compatibility)

You changed the constructor... you broke backwards compatibility.

I think this is "null" means "1" behavior is pretty sus...  It's really not very clean or intuitive. I strongly encourage you to not live this surprise.

bschofield-va: still nullable, fam

bschofield-va (approved):

PR #256, opened 2021-11-04
Api 10923 appointment transformer implementation

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-10923
Caveat: the `appointment` mapping is incomplete and will be refined at a later date TBD


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -1,52 +1,142 @@
...
 public class R4AppointmentTransformer {
   @NonNull String site;
 
   @NonNull String patientIcn;
 
   @NonNull VprGetPatientData.Response.Results rpcResults;
 
+  private Appointment.AppointmentStatus appointmentStatus(String status) {
+    switch (status) {
+      case "NO-SHOW":
```
jblefkove-va, line 32: Let's explicitly list `SCHEDULED/KEPT` and `NO ACTION TAKEN` as well

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -1,52 +1,142 @@
...
+    if (isBlank(serviceType)) {
+      return null;
+    }
+    return CodeableConcept.builder().text(serviceType).build().asList();
+  }
+
+  private Appointment toAppointment(Appointments.Appointment rpcAppointment) {
     return Appointment.builder()
         .id(idFrom(rpcAppointment.id().value()))
+        .appointmentType(appointmentType(rpcAppointment.type().name()))
```
jblefkove-va, line 125: Use the order of fields in the FHIR spec

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -1,52 +1,144 @@
...
 @Builder
+@Slf4j
 public class R4AppointmentTransformer {
   @NonNull String site;
 
   @NonNull String patientIcn;
 
   @NonNull VprGetPatientData.Response.Results rpcResults;
 
+  private Appointment.AppointmentStatus appointmentStatus(String status) {
```
jhulbert-va, line 30: What if status is null?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -1,52 +1,144 @@
...
+    switch (status) {
+      case "NO-SHOW":
+        return Appointment.AppointmentStatus.noshow;
+      case "CANCELLED":
+        return Appointment.AppointmentStatus.cancelled;
```
jhulbert-va, line 35: As a nit, can you import `AppointmentStatus` so this reads better?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -1,52 +1,144 @@
...
   }
 
-  Stream<Appointment> toFhirSkeleton() {
+  Stream<Appointment> toFhir() {
     return rpcResults
         .appointmentStream()
         .filter(Objects::nonNull)
-        .map(this::toAppointmentSkeleton)
+        .map(this::toAppointment)
         .filter(Objects::nonNull);
```
jhulbert-va, line 142: Instead of filtering twice, why not have `toAppointment` return `null` if it returns null and then filter out the nulls once at the end?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
taylorrassmannlibertyits: Ah thanks, I was accessing the `value` too quickly in my transformer. I've refactored a bit more in the incoming push.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -1,52 +1,156 @@
...
+
+    String appointmentStatus = status.value();
+
+    switch (appointmentStatus) {
+      case "NO-SHOW":
+        return AppointmentStatus.noshow;
+      case "CANCELLED":
+        return AppointmentStatus.cancelled;
+      case "SCHEDULED/KEPT":
+      case "NO ACTION TAKEN":
```
jhulbert-va, line 47: Make sure you're commenting that we are intentionally falling through here.

jhulbert-va (approved):

PR #280, opened 2021-11-17
API-11427-remove-appointment-search-date-upper-bound 

taylorrassmannlibertyits:
Removed default upper bound date for appointment search.


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4SiteAppointmentController.java
```diff
@@ -90,7 +90,7 @@ public Appointment appointmentRead(
               defaultValue = "${vista-fhir-query.default-page-size}")
           int count) {
     if (date == null || date.length == 0) {
-      date = new String[] {"ge1901", "lt2700"};
+      date = new String[] {"ge1901"};
```
jblefkove-va, line 93: Would `ge1900` also work?

taylorrassmannlibertyits: Yes, updating shortly!

jblefkove-va (approved):
:shipit:

PR #283, opened 2021-11-18
test suite project skeleton

ilaflamme-va:
- initial docker container
- steel-thread unit test


pom.xml
```diff
@@ -14,5 +14,6 @@
     <module>vista-fhir-query</module>
     <module>vista-fhir-query-mock-services</module>
     <module>vista-fhir-query-tests</module>
+    <module>vista-fhir-query-test-suite</module>
```
jhulbert-va, line 17: Not super sure about this name. At a glance, I couldn't tell you what it's meant to be/what the difference is between it and the other tests module.

vista-fhir-query-test-suite/pom.xml
```diff
@@ -0,0 +1,93 @@
...
+      <groupId>gov.va.api.lighthouse</groupId>
+      <artifactId>test-clients</artifactId>
+      <version>${oauth-java.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>gov.va.api.health</groupId>
+      <artifactId>vista-fhir-query</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
```
jhulbert-va, line 48: Do we need all of these dependencies already? For a steel-thread, it seems suspect.

jhulbert-va (approved):

PR #299, opened 2021-12-02
API-11661-field-validator

jhulbert-va:

- One thing I'm not sure about anymore is the `map` ability. Perhaps I should 86 that, thoughts?


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/FieldValidation.java
```diff
@@ -0,0 +1,58 @@
...
+
+    Function<T, ?> map();
+
+    ValidationResult<T, ?> validate(T value);
+
+    default void checkRules(SoftAssertions assertions) {
+        try {
+            assertions.assertAll();
+        } catch (AssertionError e) {
+            throw ResourceExceptions.BadRequestPayload.because(
```
bschofield-va, line 35: This isn't exactly what I had in mind, so perhaps we should chat

eclendenning-va (approved):

PR #308, opened 2021-12-03
API-11713: test tool insuranceplan read

jabando-va:


vista-fhir-query-interactive-tests/src/test/java/gov/va/api/health/vistafhirquery/interactivetests/insuranceplan/InsurancePlanReadTest.java
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
+
+public class InsurancePlanReadTest {
+
+  @Test
+  @EnabledIfSystemProperty(named = "interactive-tests", matches = "true")
+  void insurancePlanRead() {
+    TestContext ctx = new InteractiveTestContext("InsurancePlanRead");
+    ctx.read(InsurancePlan.builder().build(), ctx.property("insurancePlanId"));
```
jhulbert-va, line 15: Should that insurance plan be `null`? Feels like the interface should just not take the resource in the signature.

jabando-va: yeah I think its fine like this. The only thing its used for is building the URL, based on the resource's type `.resourceName(resourceType.getSimpleName())`

jabando-va: worth a refactor for testcontext's read() and search() so we instead get `ctx.read(InsurancePlan.class, ctx.property("insurancePlanId"))`?

jhulbert-va: I think I'd prefer that to instantiating an empty object just to find it's class name.

jhulbert-va (approved):
LGTM

bschofield-va (approved):

PR #309, opened 2021-12-06
API-11661-better-extension-exceptions

jhulbert-va:


jhulbert-va:
Resetting to update better exception hierarchy.

PR #315, opened 2021-12-07
interactive tests: organization create/update supports all test parameters

ilaflamme-va:


vista-fhir-query-interactive-tests/src/test/java/gov/va/api/health/vistafhirquery/interactivetests/organization/OrganizationUpdateTest.java
```diff
@@ -0,0 +1,594 @@
...
+
+public class OrganizationUpdateTest {
+
+  @Test
+  @EnabledIfSystemProperty(named = "interactive-tests", matches = "true")
+  void organizationUpdate() {
+    TestContext ctx = new InteractiveTestContext("OrganizationUpdate");
+    ctx.update(
+        Organization.builder()
+            .meta(Meta.builder().source(ctx.property("site")).build())
```
jhulbert-va, line 32: I don't think we need site here. The application ends up ignoring it and using the site from the url anyway.

eclendenning-va (approved):

PR #321, opened 2021-12-07
API-11661-extension-exceptions

jhulbert-va:

- This will be part 1 of a group of PR's updating exception behavior of the extension handlers
~I'm open to suggestions on the best way to represent failures in a complex extension. Right now it just acts like it's part of the top level extension group.~


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/organization/R4OrganizationToInsuranceCompanyFileTransformer.java
```diff
@@ -57,7 +57,8 @@
 
   @NonNull Organization organization;
 
-  ExtensionProcessor extensionProcessor = R4ExtensionProcessor.of(extensionHandlers());
+  ExtensionProcessor extensionProcessor =
+      R4ExtensionProcessor.of(".extension", extensionHandlers());
```
bschofield-va, line 61: CER used `.extension[]` vs. `.extension`

bschofield-va (approved):

PR #324, opened 2021-12-08
API-11661-extension-missing-required-field

jhulbert-va:

### Part 2 of ¯\_(ツ)_/¯

- This updates the extension processor to include the json path in `handle()`.
- Missing required field exception for all processors to check the `.value[x]` field.
- NestedExtension exception for the complex extension processor to give context when the failure is within it's nested `.extension[]` field.


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/RequestPayloadExceptions.java
```diff
@@ -59,7 +67,7 @@ public String getMessage() {
   public static class InvalidExtension extends InvalidField {
     @NonNull @Getter private final String definingUrl;
 
-    InvalidExtension(String jsonPath, String definingUrl, String problem) {
+    public InvalidExtension(String jsonPath, String definingUrl, String problem) {
```
bschofield-va, line 70: Does this really need to be `public`?

bschofield-va (approved):

PR #328, opened 2021-12-09
API-11661-extension-expected-at-least-one-of-fields

jhulbert-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/RequestPayloadExceptions.java
```diff
@@ -274,6 +274,19 @@ private static ExactlyOneOfFields exactlyOneOfFields(
...
 
+  public static class ExpectedAtLeastOneOfExtensionFields extends InvalidExtension {
+    @Builder
+    ExpectedAtLeastOneOfExtensionFields(
+        String jsonPath, String definingUrl, List<String> expectedAtLeastOneOfFields) {
+      super(
+          jsonPath,
+          definingUrl,
+          format(
+              "Expected at least one of fields (%s), but got neither.",
```
bschofield-va, line 285: "neither" implies only two options, but it's a list

bschofield-va (approved):

PR #334, opened 2021-12-13
inital move to only optionals, and updating of coverage

eclendenning-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/WriteableFilemanValueFactory.java
```diff
@@ -83,125 +81,105 @@
...
   }
 
   /** Build a WriteableFilemanValue from a Boolean value. */
-  public WriteableFilemanValue forRequiredBoolean(
+  public Optional<WriteableFilemanValue> forBoolean(
       @NonNull String field, int index, Boolean value, @NonNull Function<Boolean, String> mapper) {
     if (value == null) {
-      throw BadRequestPayload.because(file(), field, "Required boolean was found to be null.");
+      return Optional.empty();
     }
```
jhulbert-va, line 100: Not sure this is right. Boolean is nullable, so it stands to reason this can be given a null value (e.g. the `inForce` field in CER). I _think_ the correct behavior is to try to map and catch a broad range of exceptions before returning an optional.

jhulbert-va: I think this applies only to `Boolean` is because of the mapper.

eclendenning-va: Not sure I understand, if it is null, we arent going to add it to the fileman values, so we can just return an optional and if its required, throw an exception, if not, no harm no foul.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverageeligibilityresponse/R4CoverageEligibilityResponseToVistaFileTransformer.java
```diff
@@ -275,14 +275,14 @@ WriteableFilemanValue identifier(@NonNull Identifier identifier) {
       case "MSH-10":
         return factoryRegistry()
             .get(IivResponse.FILE_NUMBER)
-            .forRequiredIdentifier(
+            .forIdentifier(
```
jhulbert-va, line 278: This isn't returning an optional?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/WriteableFilemanValueFactory.java
ilaflamme-va: Agreed, I think this is fine as is.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverageeligibilityresponse/R4CoverageEligibilityResponseToVistaFileTransformer.java
ilaflamme-va: I would expect Optional<WriteableFilemanValue>

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/WriteableFilemanValueFactory.java
```diff
@@ -83,125 +81,105 @@
...
-    return forRequiredString(field, index, mapper.apply(value));
+    return forString(field, index, mapper.apply(value));
   }
 
   /**
    * Build a WriteableFilemanValue using the codeableConcept.coding.code field for a given system.
    */
-  public WriteableFilemanValue forRequiredCodeableConcept(
+  public Optional<WriteableFilemanValue> forCodeableConcept(
```
jhulbert-va, line 107: We should RIP this. Multi-level objects will cause generic exceptions.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceTypeFileTransformer.java
```diff
@@ -260,10 +261,13 @@ WriteableFilemanValue patientRelationshipHipaa(CodeableConcept relationship) {
...
             .get(InsuranceType.FILE_NUMBER)
-            .forRequiredString(
+            .forString(
                 InsuranceType.INSURANCE_EXPIRATION_DATE,
                 indexRegistry().get(InsuranceType.FILE_NUMBER),
-                vistaDateFormatter().format(end));
+                vistaDateFormatter().format(end)).orElseThrow(()-> RequestPayloadExceptions.InvalidConditionalField.builder()
+                        .jsonPath(".period.end")
+                        .condition("was present but not a parsable date.")
+                        .build());
```
jhulbert-va, line 270: pretty sure `format()` will throw if it can't be formatted, so this may be unnecessary. I wonder if this is worth a `forDateTime` method in the WFVF that will catch the exception and return an Optional so we can give a nicer message.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceTypeFileTransformer.java
```diff
@@ -243,10 +244,10 @@ WriteableFilemanValue patientRelationshipHipaa(CodeableConcept relationship) {
     var effectiveDate =
         factoryRegistry()
             .get(InsuranceType.FILE_NUMBER)
-            .forRequiredString(
+            .forString(
                 InsuranceType.EFFECTIVE_DATE_OF_POLICY,
                 indexRegistry().get(InsuranceType.FILE_NUMBER),
-                vistaDateFormatter().format(start));
+                vistaDateFormatter().format(start)).get();
```
jhulbert-va, line 250: I think the same thing that applies to the end date applies here.

jabando-va (dismissed):

ilaflamme-va (approved):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/R4Transformers.java
```diff
@@ -148,6 +150,25 @@ public static boolean isBlank(Object value) {
     return value == null;
   }
 
+  /** Formats an instant or returns an Empty Optional. */
+  public static Optional<String> maybeFormatDateTime(
```
jhulbert-va, line 154: I think `try` would be a better prefix than `maybe`

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/R4Transformers.java
```diff
@@ -148,6 +150,25 @@ public static boolean isBlank(Object value) {
...
+      Instant dateTime, DateTimeFormatter dateTimeFormatter) {
+    try {
+      return Optional.of(dateTimeFormatter.format(dateTime));
+    } catch (Exception e) {
```
jhulbert-va, line 158: These can be more specific. We know exactly what they'll throw.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/ResourceExceptions.java
```diff
@@ -11,25 +11,6 @@
 @SuppressWarnings("DoNotCallSuggester")
 public class ResourceExceptions {
 
-  public static class BadRequestPayload extends ResourceException {
```
jhulbert-va, line 14: image

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/WriteableFilemanValueFactory.java
```diff
@@ -43,10 +40,18 @@
...
+      return Optional.empty();
+    }
+    return forString(field, index, coding.code());
+  }
+
   /** Build a WriteableFilemanValue using the identifer.value field. */
   public Optional<WriteableFilemanValue> forIdentifier(
       String field, int index, Identifier identifier) {
-    if (identifier == null) {
+    if (identifier == null || isBlank(identifier.value())) {
```
jhulbert-va, line 54: Why we add this back?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceTypeFileTransformer.java
```diff
@@ -244,28 +241,57 @@ WriteableFilemanValue patientRelationshipHipaa(CodeableConcept relationship) {
       throw MissingRequiredField.builder().jsonPath(".period.start").build();
     }
     List<WriteableFilemanValue> dates = new ArrayList<>(2);
-    var start = Instant.parse(period.start());
+    var startInstant =
+        maybeParseDateTime(period.start())
+            .orElseThrow(
+                () ->
+                    RequestPayloadExceptions.UnexpectedValueForField.builder()
```
jhulbert-va, line 248: Better import plz

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/extensionprocessing/DateTimeExtensionHandler.java
```diff
@@ -47,17 +47,28 @@ public static DateTimeExtensionHandlerBuilder forDefiningUrl(String definingUrl)
...
+        maybeFormatDateTime(dateTime, dateTimeFormatter)
+            .orElseThrow(
+                () ->
+                    UnexpectedValueForExtensionField.builder()
+                        .jsonPath(jsonPath)
+                        .definingUrl(definingUrl())
+                        .dataType("http://hl7.org/fhir/R4/datatypes.html#dateTime")
+                        .valueReceived(extension.valueDateTime())
+                        .build());
+    ;
```
jhulbert-va, line 71: Why not `flatMap` instead of separating it and then throwing the same exception?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/witnessprotection/WitnessProtectionAdvice.java
```diff
@@ -170,7 +170,7 @@ public Object handleEmptyBody(
       @org.springframework.lang.NonNull MethodParameter methodParameter,
       @org.springframework.lang.NonNull Type type,
       @org.springframework.lang.NonNull Class<? extends HttpMessageConverter<?>> converterType) {
-    throw BadRequestPayload.because("Request payload is empty.");
+    throw new RequestPayloadExceptions.EmptyRequestPayload();
```
jhulbert-va, line 173: import

jhulbert-va (approved):

PR #342, opened 2021-12-21
problem condition transformer

eclendenning-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/condition/VistaProblemToR4ConditionTransformer.java
```diff
@@ -0,0 +1,179 @@
...
+    return toResourceId(patientIcn, site, VprGetPatientData.Domains.problems, id);
+  }
+
+  private Condition toCondition(Problems.Problem rpcCondition) {
+    if (isBlank(rpcCondition)) {
+      return null;
+    }
+    return Condition.builder()
+        .id(idFrom(rpcCondition.id().value()))
+        .meta(Meta.builder().source(site).build())
```
jabando-va, line 136: .lastUpdated | updated | Direct relationship converted from fileman date
-- | -- | --

I think this goes here?

jabando-va (approved):

PR #348, opened 2022-01-10
API-12194: new Condition per Visit.Icd

jabando-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/condition/VprVisitToR4ConditionTransformer.java
```diff
@@ -43,13 +44,16 @@
         .asList();
   }
 
-  CodeableConcept code(List<Visits.Icd> icds) {
+  CodeableConcept code(Visits.Icd icds) {
     List<Coding> codings =
-        Safe.stream(icds).map(this::icdToCoding).filter(Objects::nonNull).collect(toList());
+        Safe.stream(List.of(icds))
```
bschofield-va, line 49: This doesn't make sense to me ... we have 1 Icd, which we make into a list of one, which we make into a _safe_ stream of 1.

Safe stream is mean to deal with potentially null collections, which we _know_ is not the case since you construct the list explicitly.

You can make a stream of fixed size with `Stream.of(a,b,c)` but why? You have 1 thing and a method that processes 1 thing, just skip the stream altogether.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/condition/VprVisitToR4ConditionTransformer.java
```diff
@@ -44,13 +44,11 @@
         .asList();
   }
 
-  CodeableConcept code(List<Visits.Icd> icds) {
-    List<Coding> codings =
-        Safe.stream(icds).map(this::icdToCoding).filter(Objects::nonNull).collect(toList());
-    return CodeableConcept.builder().coding(codings).build();
+  CodeableConcept code(Visits.Icd icd) {
+    return CodeableConcept.builder().coding(icdToCoding(icd).asList()).build();
```
bschofield-va, line 48: If `icd` is null, this will NPE since `asList()` will be invoked against a `null` return. Probably should not even create the CodeableConcept if `icd` is `null`.

bschofield-va (approved):

PR #354, opened 2022-01-17
API-11859: Condition clinical status search support

jabando-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/condition/R4ConditionCollector.java
```diff
@@ -31,7 +33,15 @@
                         .site(site)
                         .vistaProblem(problem)
                         .build()
-                        .toFhir());
+                        .toFhir())
+            .filter(
```
jhulbert-va, line 37: I think if we're going to filter by clinical-status, we should be filtering _all_ results by clinical-status.

jhulbert-va (approved):

PR #359, opened 2022-01-19
API-11861: Condition onset date search

jabando-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/condition/R4ConditionCollector.java
```diff
@@ -71,6 +77,23 @@ private boolean conditionHasRequestedCode(Condition condition) {
...
+      return false;
+    }
+    if (isBlank(onsetDate().start())) {
+      return onsetDate().stop().isAfter(Instant.parse(condition.onsetDateTime()));
+    }
+    if (isBlank(onsetDate().stop())) {
+      return onsetDate().start().isBefore(Instant.parse(condition.onsetDateTime()));
+    }
+    return onsetDate().start().isBefore(Instant.parse(condition.onsetDateTime()))
+        && onsetDate().stop().isAfter(Instant.parse(condition.onsetDateTime()));
```
jhulbert-va, line 94: Just to reflect my comment on the PR as well; I think this logic should move to `DateSearchBoundaries` as a `test(Instant)` or `check(Instant)` method. This will likely become common place for date searching by specific fields in a resource.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/DateSearchBoundaries.java
```diff
@@ -202,6 +204,17 @@ private void invalidDateCombination() {
         "Bad date search combination : date=" + date1.toString() + "&" + date2.toString());
   }
 
+  /** Stuff. */
+  public boolean isDateWithinBounds(Instant date) {
+    if (isBlank(start())) {
```
jhulbert-va, line 209: `start` and `stop` are `Instant`'s. I think just a null check should be fine.

jhulbert-va: These also assume that if start is null that stop isnt, and I'm not sure that's a safe assumption to make.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/condition/R4ConditionCollector.java
```diff
@@ -51,6 +56,7 @@
             : Arrays.stream(clinicalStatusCsv.split(",", -1)).collect(toSet());
     this.code = code;
     this.results = results;
+    this.onsetDate = isBlank(date) ? null : DateSearchBoundaries.of(date);
```
jhulbert-va, line 59: Date search boundaries already checks for null.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/DateSearchBoundaries.java
```diff
@@ -202,6 +203,20 @@ private void invalidDateCombination() {
         "Bad date search combination : date=" + date1.toString() + "&" + date2.toString());
   }
 
+  /** Check if a given instant is within the bounds of the start and stop. */
+  public boolean isDateWithinBounds(Instant date) {
+    if ((start() == null && stop() == null) || date == null) {
+      return false;
+    }
```
jhulbert-va, line 210: Can we make this two checks? First, if the value given is null, return false. Then, if both start and stop are null, return false. One checks the value given is valid, the other checks the boundaries are valid. I think this will make the code cleaner and easier to understand.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/condition/R4ConditionCollector.java
```diff
@@ -71,6 +77,16 @@ private boolean conditionHasRequestedCode(Condition condition) {
     return Safe.stream(condition.code().coding()).anyMatch(coding -> code().equals(coding.code()));
   }
 
+  private boolean conditionHasRequestedOnsetDate(Condition condition) {
+    if (isBlank(onsetDate())) {
+      return true;
+    }
+    if (isBlank(condition.onsetDateTime())) {
+      return false;
+    }
```
jhulbert-va, line 86: You added this check to `DateSearchBoundaries`.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/condition/R4ConditionCollector.java
```diff
@@ -71,6 +77,16 @@ private boolean conditionHasRequestedCode(Condition condition) {
...
+    return onsetDate().isDateWithinBounds(Instant.parse(condition.onsetDateTime()));
```
jhulbert-va, line 87: Two things:
1. Instead of parsing here, overload the `isDateWithinBounds` to take a string parameter and parse it there.
2. Use the `FhirDateTime` class to parse the date.

jhulbert-va (approved):

PR #361, opened 2022-01-20
initial transformer

eclendenning-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -0,0 +1,46 @@
...
+@Value
+public class R4CoverageToInsuranceBufferTransformer {
+  FilemanFactoryRegistry factoryRegistry;
+
+  FilemanIndexRegistry indexRegistry;
+
+  @NonNull Coverage coverage;
+
+  @Builder
+  R4CoverageToInsuranceBufferTransformer(@NonNull Coverage coverage, ZoneId timezone) {
```
jhulbert-va, line 24: Don't need zoneId yet.

PR #367, opened 2022-01-26
obs serach bad request when no query params provided

ilaflamme-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/observation/R4SiteObservationController.java
```diff
@@ -100,6 +102,10 @@ public Observation observationRead(
               required = false,
               defaultValue = "${vista-fhir-query.default-page-size}")
           int count) {
+    if (allBlank(patient, id, identifier, category, code, date)) {
+      throw BadSearchParameters.because(
+          "Observation search requires query parameters. None were provided.");
```
jhulbert-va, line 107: Should this say something like requires at least one of patient, _id, or identifier?

jhulbert-va (approved):
It's only a nit, so if more than one person disagrees, I say smerge the heck out of this.

bschofield-va (approved):

PR #374, opened 2022-01-28
coverage read from insurance buffer

eclendenning-va:


vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4SiteInsuranceBufferCoverageControllerTest.java
```diff
@@ -93,6 +122,17 @@ void coverageCreateUsingInsuranceTypeControllerHack() {
 
   @Test
   void coverageRead() {
```
jhulbert-va, line 124: We'll probably want a test that has it going through the OG controller as well.

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4SiteInsuranceBufferCoverageControllerTest.java
```diff
@@ -39,6 +42,24 @@ private R4BundlerFactory _bundlerFactory() {
         .build();
   }
 
+  private R4SiteInsuranceBufferCoverageController _controller() {
```
jhulbert-va, line 45: `_insuranceBufferCoverageController()` is already defined.

jhulbert-va (approved):

PR #379, opened 2022-01-31
API-13096-insplan-fields

jhulbert-va:
# [API-13096](https://vajira.max.gov/browse/API-13096)


jhulbert-va:
Mappings changed and code will need to be refactored.

PR #380, opened 2022-01-31
initial contained resource reader

eclendenning-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/ContainedResourceReader.java
```diff
@@ -0,0 +1,32 @@
...
+
+  public ContainedResourceReader(@NonNull List<Resource> contained) {
+    this.containedResources = contained.stream().collect(toMap(Resource::id, Function.identity()));
+  }
+
+  /** Get resource if contained resources have the specified reference. */
+  public <T> Optional<T> find(Class<T> type, String id) {
+    Resource containedResource = containedResources.get(id);
+    if (isBlank(containedResource)) {
+      throw InvalidReferenceId.builder().jsonPath(".contained[]").build();
```
jhulbert-va, line 25: I don't think this is right. It's returning an optional. If we can't find the resource, it should return an empty optional. The transformer would then make the decision on what to throw.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/ContainedResourceReader.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+    if (type.isInstance(containedResource)) {
+      return Optional.of(type.cast(containedResource));
```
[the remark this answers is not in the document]
ilaflamme-va: I agree, similar to the throw above.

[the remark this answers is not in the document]
eclendenning-va:
```public T cast(Object obj) {
        if (obj != null && !isInstance(obj))
            throw new ClassCastException(cannotCastMsg(obj));
        return (T) obj;
    }```
    
    we're already catching the possibility of it throwing, and based off of your other comment, returning Optional.empty and allowing the transformer to make a nice exception

File: vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java [hunk 3]
```diff
@@ -55,6 +70,34 @@ WriteableFilemanValue dateEntered() {
...
+    return factoryRegistry()
+        .get(InsuranceVerificationProcessor.FILE_NUMBER)
+        .forString(
+            InsuranceVerificationProcessor.GROUP_NAME,
+            indexRegistry().get(InsuranceVerificationProcessor.FILE_NUMBER),
+            name)
+        .orElse(null);
+  }
+
+  WriteableFilemanValue groupNumber(InsurancePlan containedInsurancePlan) {
```
jhulbert-va, line 83: I think this should just take in the list of identifiers.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -55,6 +70,34 @@ WriteableFilemanValue dateEntered() {
...
+    Optional<Identifier> maybeGroupNumberIdentifier =
+        containedInsurancePlan.identifier().stream()
+            .filter(i -> InsurancePlanStructureDefinitions.GROUP_NUMBER.equals(i.system()))
+            .findFirst();
```
jhulbert-va, line 87: We should check the size of this.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -88,6 +131,47 @@ WriteableFilemanValue insuranceCompanyName() {
...
+            .filter(
+                c -> {
+                  if (c.type() == null) {
+                    return false;
+                  }
+                  return c.type().coding().stream()
+                      .anyMatch(
+                          coding ->
+                              COVERAGE_CLASS_CODE_SYSTEM.equals(coding.system())
+                                  && "group".equals(coding.code()));
```
jhulbert-va, line 149: Should we abstract this to a method? `isGroupPlan` or something like that?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/ContainedResourceReader.java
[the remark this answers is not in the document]
jhulbert-va: I'd be cool returning an optional empty. I guess our exception will be something like `ContainedResourceNotFound`?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/ContainedResourceReader.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class ContainedResourceReader {
+  Map<String, Resource> containedResources;
+
+  public ContainedResourceReader(@NonNull List<Resource> contained) {
+    this.containedResources = contained.stream().collect(toMap(Resource::id, Function.identity()));
+  }
+
+  /** Get resource if contained resources have the specified reference. */
+  public <T> Optional<T> find(Class<T> type, String id) {
```
jhulbert-va, line 21: We should `@NonNull` or perform null checks on these bad bois.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/ContainedResourceReader.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+import lombok.NonNull;
+
+public class ContainedResourceReader {
+  Map<String, Resource> containedResources;
+
+  public ContainedResourceReader(@NonNull List<Resource> contained) {
+    this.containedResources = contained.stream().collect(toMap(Resource::id, Function.identity()));
+  }
```
jhulbert-va, line 18: Instead of `@NonNull` why not just `Safe.stream()` it so that we can always create a reader (even if there are no contained resources).

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -88,6 +139,48 @@ WriteableFilemanValue insuranceCompanyName() {
         .get();
   }
 
+  List<WriteableFilemanValue> insurancePlan() {
+    if (isBlank(coverage.coverageClass()) || isBlank(coverage.contained())) {
```
jhulbert-va, line 143: Should these occur in different locations? We need to check `.class` first and throw if there isn't one (groupNumber is required). Then we'll throw if we can't find the InsurancePlan with the id specified.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -55,6 +70,42 @@ WriteableFilemanValue dateEntered() {
...
+    if (groupNumbers.size() != 1) {
+      throw UnexpectedNumberOfValues.builder()
+          .exactExpectedCount(1)
+          .receivedCount(groupNumbers.size())
+          .jsonPath(".contained[].identifiers[]")
+          .build();
+    }
+    Optional<Identifier> maybeGroupNumberIdentifier = groupNumbers.stream().findFirst();
+    return maybeGroupNumberIdentifier
+        .map(
```
jhulbert-va, line 97: `.flatMap` this bad boi!

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -45,6 +55,10 @@
             .withZone(timezone == null ? ZoneId.of("UTC") : timezone);
   }
 
+  private ContainedResourceReader containedResourceReader() {
+    return new ContainedResourceReader(coverage.contained());
+  }
+
```
jhulbert-va, line 61: Should this happen as part of the constructor?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -55,6 +69,41 @@ WriteableFilemanValue dateEntered() {
...
+            .toList();
+    if (groupNumbers.size() != 1) {
+      throw UnexpectedNumberOfValues.builder()
+          .exactExpectedCount(1)
+          .receivedCount(groupNumbers.size())
+          .jsonPath(".contained[].identifiers[]")
+          .build();
+    }
+    Optional<Identifier> maybeGroupNumberIdentifier = groupNumbers.stream().findFirst();
+    return maybeGroupNumberIdentifier
```
jhulbert-va, line 95: Should we just use `.get(0)` instead?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -88,6 +137,51 @@ WriteableFilemanValue insuranceCompanyName() {
...
+          .build();
+    }
+    String referenceId = filteredCoverageTypes.get(0).value();
+    InsurancePlan containedInsurancePlan =
+        containedResourceReader()
+            .find(InsurancePlan.class, referenceId)
+            .orElseThrow(
+                () ->
+                    RequestPayloadExceptions.InvalidReferenceId.builder()
+                        .jsonPath(".contained[]")
```
jhulbert-va, line 165: I think this is `.coverageClass[0].value`

jhulbert-va (dismissed):

jhulbert-va (dismissed):

jhulbert-va (approved):
I think we should be checking those failure cases that return empty optionals, but I think future PR that.

PR #384, opened 2022-02-02
415 response for HttpMediaTypeNotSupported

ilaflamme-va:


jabando-va (approved):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/WebExceptionHandler.java
```diff
@@ -311,6 +310,13 @@ OperationOutcome handleUnprocessableEntity(Exception e, HttpServletRequest reque
     return responseFor("structure", e, request, emptyList(), true, null);
   }
 
+  @ExceptionHandler(HttpMediaTypeNotSupportedException.class)
+  @ResponseStatus(HttpStatus.UNSUPPORTED_MEDIA_TYPE)
+  OperationOutcome handleUnsupportedMediaType(
+      HttpMediaTypeNotSupportedException e, HttpServletRequest request) {
+    return responseFor("unsupported-media-type", e, request, emptyList(), true, null);
```
bschofield-va, line 317: I think this is the wrong code... 
http://hl7.org/fhir/R4/valueset-issue-type.html

bschofield-va: `internal-server-error` is also wrong

ilaflamme-va: Yep, should be `exception`

bschofield-va (approved):

PR #385, opened 2022-02-04
Steel thread med dispense

msilverman-va:
Adds API, controller, lean transformer, header advice, witness protection, and integration tests for medication dispense.

Co-authored-by: Joshua Miller <jmiller@rise8.us>
Co-authored-by: Thomas Reynolds <treynolds@rise8.us>
Co-authored-by: Michael Silverman <msilverman@rise8.us>


vista-fhir-query-tests/src/main/java/gov/va/api/health/vistafhirquery/tests/TestIds.java
```diff
@@ -22,6 +22,8 @@
 
   @NonNull String insurancePlan;
 
+  @NonNull String meds;
```
bschofield-va, line 25: `private static TestIds idsForProductionEnvironment()` does not initialize this property, NPE guaranteed

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+  public Stream<MedicationDispense> toFhir() {
+    return rpcResults.medStream().map(this::toMedicationDispense).filter(Objects::nonNull);
+  }
+
+  private MedicationDispense toMedicationDispense(Meds.Med rpcMed) {
+    if (rpcMed == null) {
+      return null;
+    }
+    return MedicationDispense.builder()
+        .id(idFrom(rpcMed.id().value()))
```
bschofield-va, line 43: potential NPE, see Observation and Condition transformers for a better example on  how to do this

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseResponseIncludesIcnHeaderAdvice.java
```diff
@@ -0,0 +1,42 @@
...
+      @Autowired AlternatePatientIds alternatePatientIds) {
+    delegate =
+        ResponseIncludesIcnHeaderAdvice.<MedicationDispense, Bundle>builder()
+            .type(MedicationDispense.class)
+            .bundleType(Bundle.class)
+            .extractResources(bundle -> bundle.entry().stream().map(AbstractEntry::resource))
+            .extractIcns(
+                resource ->
+                    referenceIdFromUri(resource.subject())
+                        .filter(Objects::nonNull)
```
jhulbert-va, line 37: `referenceIdFromUri` returns an `Optional`, so you shouldn't need to filter by NonNull. `Optional.map` will only perform an action if the contained value is present.

vista-fhir-query-tests/src/main/java/gov/va/api/health/vistafhirquery/tests/TestIds.java
jnmiller-va: @bschofield-va could we unpack this a bit?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -0,0 +1,48 @@
...
+  String idFrom(ValueOnlyXmlAttribute maybeId) {
+    String id = valueOfValueOnlyXmlAttribute(maybeId);
+    if (isBlank(id)) {
+      return null;
+    }
+    return toResourceId(patientIcn, site, VprGetPatientData.Domains.meds, id);
+  }
+
+  public Stream<MedicationDispense> toFhir() {
+    return rpcResults.medStream().filter(Objects::nonNull).map(this::toMedicationDispense);
```
jhulbert-va, line 37: With vista, I believe this should be checking the object isn't empty, not just null. 
https://github.com/department-of-veterans-affairs/lighthouse-charon/blob/97918c4441e0e342d1e0ddee6f08ef26d2f410ce/charon-models/src/main/java/gov/va/api/lighthouse/charon/models/vprgetpatientdata/Meds.java#L130

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseWitnessProtectionAgent.java
```diff
@@ -0,0 +1,27 @@
...
+
+  @NonNull private final ProtectedReferenceFactory protectedReferenceFactory;
+
+  @Override
+  public Stream<ProtectedReference> referencesOf(MedicationDispense resource) {
+    return Stream.concat(
+        Stream.of(protectedReferenceFactory.forResource(resource, resource::id)),
+        Stream.of(
+            protectedReferenceFactory.forReferenceWithoutSite(resource.subject()).orElse(null)));
+  }
```
jhulbert-va, line 26: A single `Stream.of()` should work here I think. No need to concat.

vista-fhir-query-tests/src/main/java/gov/va/api/health/vistafhirquery/tests/TestIds.java
jhulbert-va: Lomboks `@NonNull` annotation will throw an NPE if a field annotated with it is null. 
In `SystemDefinitions`, we're initializing three sets of TestIds; local, synthetic, and production. That being said, in this PR, only the local and synthetic ids (not the production ids - `idsForProductionEnvironment`) are being set, so when this tries to run integration tests against the staging/production environments, the tests will throw an NPE.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseResponseIncludesIcnHeaderAdvice.java
kivarsson-va: that was our thinking as well, @bschofield-va curious why you added the filter?

vista-fhir-query-tests/src/main/java/gov/va/api/health/vistafhirquery/tests/TestIds.java
jnmiller-va: Should we just add a TODO like the others?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
kivarsson-va: yeah we can change it to an `isBlank()` check, some of the transformers do that and some filter on nonNull so we weren't sure which was a better pattern

jhulbert-va: We've found that vista will sometimes return empty objects, so that's definitely a check we need to perform.

vista-fhir-query-tests/src/main/java/gov/va/api/health/vistafhirquery/tests/TestIds.java
jhulbert-va: Yeah. The comment would just be `ToDo` like the others. Unfortunately, we would need a production test patient that spans both vista and MPI. We have yet to figure out a way to find a patient that does that and has the data we need.

jnmiller-va: That's what I was thinking! Interesting problem for sure.

PR #398, opened 2022-02-10
Api 13404 rpc principals ignores unknown fields

ilaflamme-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/config/RpcPrincipalConfig.java
```diff
@@ -25,7 +26,9 @@
   RpcPrincipalLookup loadPrincipals(
       @NonNull @Value("${vista-fhir-query.rpc-principals.file}") String principalsFile) {
     RpcPrincipals rpcPrincipals =
-        new ObjectMapper().readValue(new File(principalsFile), RpcPrincipals.class);
+        new ObjectMapper()
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
jhulbert-va, line 30: Should we have a ToDo/task to track the eventual removal of this?

vista-fhir-query/src/test/resources/principals.json
```diff
@@ -6,6 +6,10 @@
         "NACHO",
         "BURRITO"
       ],
+      "whoDis?" : [
```
bschofield-va, line 9: I think this is confusing ... the name implies these are prinicpals. But really, they are `principals-with-unknown-values.json` crafted specifically for a certain test.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/config/RpcPrincipalConfig.java
bschofield-va: This entire class will be deleted when Charon V0 goes away. There is a `RpcPrincipalV1Config` that also exists and is not lenient.

bschofield-va (approved):

PR #399, opened 2022-02-10
API-13096: Insurance buffer insurance plan fields

jabando-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/InsuranceBufferToR4CoverageTransformer.java
```diff
@@ -126,6 +138,66 @@ private Reference beneficiary(@NonNull String patientIcn, Optional<String> maybe
...
+                InsuranceBufferStructureDefinitions.BANKING_IDENTIFICATION_NUMBER),
+            insurancePlanIdentifier(
+                entry,
+                InsuranceVerificationProcessor.PROCESSOR_CONTROL_NUMBER_PCN,
+                InsuranceBufferStructureDefinitions.PROCESSOR_CONTROL_NUMBER_PCN))
+        .filter(Objects::nonNull)
+        .toList();
+  }
+
+  private List<InsurancePlan.Plan> plan(FilemanEntry entry) {
```
jhulbert-va, line 189: Can we name this `insurancePlanType`? `.type(type())` reads weird below.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/InsuranceBufferToR4CoverageTransformer.java
```diff
@@ -126,6 +138,66 @@ private Reference beneficiary(@NonNull String patientIcn, Optional<String> maybe
...
+    CodeableConcept typeOfPlan =
+        type(
+            entry,
+            InsuranceVerificationProcessor.TYPE_OF_PLAN,
+            InsuranceBufferStructureDefinitions.TYPE_OF_PLAN);
+    if (typeOfPlan == null) {
+      return emptyList();
+    }
+    return List.of(InsurancePlan.Plan.builder().type(typeOfPlan).build());
```
jhulbert-va, line 198: `.asList()`

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -76,11 +88,10 @@ WriteableFilemanValue groupName(String name) {
...
   }
 
-  WriteableFilemanValue groupNumber(List<Identifier> identifiers) {
+  WriteableFilemanValue identifier(
+      List<Identifier> identifiers, String fieldNumber, String system) {
     List<Identifier> groupNumbers =
-        identifiers.stream()
-            .filter(i -> InsuranceBufferStructureDefinitions.GROUP_NUMBER.equals(i.system()))
-            .toList();
+        identifiers.stream().filter(i -> system.equals(i.system())).toList();
```
jhulbert-va, line 94: `Safe.stream(identifiers)`

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -76,11 +88,10 @@ WriteableFilemanValue groupName(String name) {
         .orElse(null);
   }
 
-  WriteableFilemanValue groupNumber(List<Identifier> identifiers) {
+  WriteableFilemanValue identifier(
```
jhulbert-va, line 91: Feels like this could be smarter. We're iterating this list of identifiers 3 times.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -155,13 +163,71 @@ WriteableFilemanValue insuranceCompanyName() {
...
+                    InsuranceVerificationProcessor.PROCESSOR_CONTROL_NUMBER_PCN,
+                    InsuranceBufferStructureDefinitions.PROCESSOR_CONTROL_NUMBER_PCN),
+                typeOfPlan(containedInsurancePlan.plan())),
+            insurancePlanExtensionProcessor().process(containedInsurancePlan.extension()).stream())
         .filter(Objects::nonNull)
         .toList();
   }
 
+  private List<ExtensionHandler> insurancePlanExtensionHandlers() {
+    return List.of(
```
jhulbert-va, line 188: Looking at the insurance buffer mappings, none of these are required.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -155,13 +163,71 @@ WriteableFilemanValue insuranceCompanyName() {
...
+                    InsuranceVerificationProcessor.GROUP_NUMBER,
+                    InsuranceBufferStructureDefinitions.GROUP_NUMBER),
+                identifier(
+                    containedInsurancePlan.identifier(),
+                    InsuranceVerificationProcessor.BANKING_IDENTIFICATION_NUMBER,
+                    InsuranceBufferStructureDefinitions.BANKING_IDENTIFICATION_NUMBER),
+                identifier(
+                    containedInsurancePlan.identifier(),
+                    InsuranceVerificationProcessor.PROCESSOR_CONTROL_NUMBER_PCN,
+                    InsuranceBufferStructureDefinitions.PROCESSOR_CONTROL_NUMBER_PCN),
```
jhulbert-va, line 180: Group Number is required, but the other two aren't.

bschofield-va (changes requested):
blocking merge

bschofield-va (approved):

jabando-va:
Currently adding insuranceplan fields to coverageIT

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/IdentifierReader.java
```diff
@@ -0,0 +1,71 @@
...
+    this.indexRegistry = indexRegistry;
+    Map<String, Identifier> identifiersBySystem = new HashMap<>(identifiers.size());
+    identifiers.forEach(
+        identifier -> {
+          if (identifiersBySystem.putIfAbsent(identifier.system(), identifier) != null) {
+            throw UnexpectedNumberOfValues.builder()
+                .exactExpectedCount(1)
+                .receivedCount(2)
+                .jsonPath(".contained[].identifiers[]")
+                .build();
```
jhulbert-va, line 43: I think we need to do better with this exception. It gives no insight into which identifier is the problem. It also assumes there's only two. PutIfAbsent will return the same if there are any more than one of an identifier, not just 2. 

Perhaps we also need a way to define what contained resource is the problem. In this case it could be an identifier for any one of 3 resources.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/IdentifierReader.java
```diff
@@ -0,0 +1,71 @@
...
+        });
+    this.identifiersBySystem = identifiersBySystem;
+  }
+
+  /** IdentifierReader processes a given collection of the expected identifier records. */
+  public List<WriteableFilemanValue> process(Collection<IdentifierRecord> records) {
+    return Safe.stream(records)
+        .map(
+            record -> {
+              if (record.isRequired() && !identifiersBySystem().containsKey(record.system())) {
```
jhulbert-va, line 54: I think we should try to return fast here.
```
if(!identifierExists) {
   if (required ) { throw } 
  else {return null}
}
return filemanFactory.shenanigans().orElse(null);

```

ilaflamme-va (approved):

PR #401, opened 2022-02-11
search by tag

ilaflamme-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/api/R4EndpointApi.java
```diff
@@ -160,5 +160,10 @@ Endpoint endpointRead(
               in = ParameterIn.QUERY,
               name = "status",
               description = "The availability of the endpoint for use.")
-          String status);
+          String status,
+      @Parameter(
+              in = ParameterIn.QUERY,
+              name = "tag",
+              description = "The custom tag used to indentify the endpoint.")
```
jhulbert-va, line 167: I don't like the use of `custom` here. I think something like 
```
A tag used that can be used to identify the endpoints usage.
```
That way we don't open the possibility of some end user asking us to make them a custom tag.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/endpoint/R4EndpointController.java
```diff
@@ -77,16 +77,22 @@ public Endpoint endpointRead(@PathVariable("site") String site) {
...
           int count,
       @RequestParam(name = "patient", required = false) String patient,
-      @RequestParam(value = "status", required = false) String status) {
+      @RequestParam(value = "status", required = false) String status,
+      @RequestParam(name = "tag", required = false) String tag) {
     if (!isSupportedStatus(status)) {
       return toBundle(request).apply(emptySet());
     }
-    Set<String> stations = stations(SUPPORTED_RPC);
+    Set<String> stations = stationsByRpcName(SUPPORTED_RPC);
```
jhulbert-va, line 85: Wasn't the point of the tags to get rid of this boi? The initial set should be all available sites now, right?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/api/R4EndpointApi.java
ilaflamme-va: words are hard.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/endpoint/R4EndpointController.java
ilaflamme-va: Ah, you may be right.

ilaflamme-va: Will remove this after the fact, otherwise we cause breakage. Must maintain backwards compatibility.

bschofield-va (approved):

PR #402, opened 2022-02-11
API-000: Medication Dispense Transformer Dosage

msilverman-va:
Add dosage instruction to medication dispense    

Co-authored-by: Joshua Miller <jmiller@rise8.us>


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -74,6 +75,14 @@ private Reference destination(ValueOnlyXmlAttribute routing) {
     return Reference.builder().display(display).build();
   }
 
+  private List<Dosage> dosageInstruction(String sig, String ptInstructions) {
+    return Dosage.builder()
```
bschofield-va, line 79: We should not emit an empty `Dosage`. If you can't populate a useful structure, do nothing.

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformerTest.java
```diff
@@ -82,4 +82,13 @@ void toFhirRpcMedWithMultipleFillsAndReleaseDate() {
         _transform(MedicationDispenseSamples.Vista.create().resultsWithMultipleFills());
     assertThat(dispenses.count()).isEqualTo(2);
   }
+
+  @Test
+  void toFhir_no_dosage_instructions() {
```
bschofield-va, line 87: please use our naming conventions for tests

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -74,6 +76,17 @@ private Reference destination(ValueOnlyXmlAttribute routing) {
     return Reference.builder().display(display).build();
   }
 
+  private List<Dosage> dosageInstruction(String sig, String ptInstructions) {
+    if (isBlank(sig) && isBlank(ptInstructions)) {
+      return null;
+    }
+    return Dosage.builder()
+        .text(Objects.toString(sig, ""))
```
bschofield-va, line 84: You've already established these values are not null... no need for `Objects.toString`

kivarsson-va: the check was if they're both null to not return an empty dosage. one or the other could still be empty

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/MedicationDispenseSamples.java
```diff
@@ -182,6 +195,14 @@ public MedicationDispense medicationDispenseInProgress() {
       return results().meds(Meds.builder().medResults(List.of(medWithMultipleFills())).build());
     }
 
+    public VprGetPatientData.Response.Results resultsWithNoPatientInstructions() {
+      return results(med().ptInstructions(null).sig(null));
+    }
+
+    public VprGetPatientData.Response.Results resultsWithNoPatientInstructions2() {
```
bschofield-va, line 202: I don't think this is a sustainable approach to add _every_ sample a particular test might need to class, especially when named so ambiguously...

A slightly better approach would be to move these _customized_ sample values to the test that needs them.

A 10/10 approach would be to make the specific transformer methods `package` protected and test those specifically.

```
static Stream<Argument> dosageInstruction() {
   return Stream.of(
       Argument.of(null,null,null),
       Argument.of("","",null),
       Argument.of("a","b",Dosage.builder()....whatever...asList(),
       ... more cases here ...
       );
}

@ParameterizedTest @MethdSource void dosageInstruction(String sig, String ptInstructions, List<Dosage> expected) {
  var actual = _transormer().dosageInstruction(sig, ptInstructions);
   assertThat(actual).isEqualTo(expected)
}
```

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
bschofield-va: right .. right, sorry, still do not need emit `""` values, e.g. `"text": ""` is less good than `< text not emitted >`

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -74,6 +76,17 @@ private Reference destination(ValueOnlyXmlAttribute routing) {
     return Reference.builder().display(display).build();
   }
 
+  private List<Dosage> dosageInstruction(String sig, String ptInstructions) {
+    if (isBlank(sig) && isBlank(ptInstructions)) {
```
jhulbert-va, line 80: The `R4Transformers` class also has an `allBlank()`that can accomplish this.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
bschofield-va: Sorry... that ^ might not be clear...

Should not do this:
```
  {  "text": "", "patientInstruction": "stuff" }
```
Instead do this:
```
  {  "patientInstruction": "stuff" }
```

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/MedicationDispenseSamples.java
kivarsson-va: we don't test many different cases for each field so the parameterized tests might be a little verbose. agree the resources for specific tests can be moved into the respective tests, though 👌

bschofield-va: we do in other places, and this transformer will ultimately be upgraded to so, too.
in general, I prefer to test the smaller transformations specifically vs. using something big like `toFhir()` to verify a small change to some nested object,

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -149,7 +162,9 @@ private String toDate(String date) {
     return R4Transformers.optionalInstantToString(toHumanDateTime(date));
   }
 
-  /** Convert VistA Med data to R4 Fhir data. */
+  /**
+   * Convert VistA Med data to R4 Fhir data.
+   */
```
bschofield-va, line 167: while you're here.... can you make `med.fill().stream()` be `Safe.stream(med.fill())`?

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/MedicationDispenseSamples.java
kivarsson-va: we were trying to avoid having to make them all public, but if y'all would prefer us to do it that way we can do that

bschofield-va (changes requested):
blocking merge

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/MedicationDispenseSamples.java
bschofield-va: I refactored the tests to have more focus. `package` protection is a reasonable compromise between `public` and `private`_-but-not-easily-testable_ protection.

bschofield-va (approved):

PR #403, opened 2022-02-11
API 13157 Populate Participant Location Appointment

agarcia-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -82,12 +83,23 @@ String idFrom(String id) {
...
-  List<Appointment.Participant> participants() {
+  Appointment.Participant location(CodeAndNameXmlAttribute facility) {
+    if (facility == null || facility.code() == null) {
+      return null;
+    }
+    var facilityIen = facility.code();
+    var display = "vha_" + site + "_" + facilityIen;
+    return Appointment.Participant.builder()
+        .actor(Reference.builder().display(display).build())
+        .status(Appointment.ParticipationStatus.accepted)
```
taylorrassmannlibertyits, line 94: Are we sure this field needs to be populated?

agarcia-va: Yes, status is defined as not null. Not doing so causes IT failures. I went with the assumption that if the Appointment is registered, the location would already need to be confirmed, so `accepted` was appropriate.

ilaflamme-va: It's dangerous to _assume_ these kind of things, as developers. I highly recommend confirming with KBS (or whoever your source of truth is).

agarcia-va: @ilaflamme-va Noted, I will re-check this. I did not see any case where location was saved as anything else, but I will confirm this is the case and report again in this thread.

agarcia-va: @ilaflamme-va I verified from the `lighthouse-charon` [wiki's mapping notes](https://github.com/department-of-veterans-affairs/lighthouse-charon/wiki/Incubator:-VPR-GET-PATIENT-DATA-(Appointment)-to-R4-Appointment#patient-participant). It does not specify about `location`, but `Appointment.Participant`'s mapping does default to `accepted` in its `status`, so the same should hold true for `location`. Both would be participants, after all.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -146,6 +158,7 @@ private Appointment toAppointment(Appointments.Appointment rpcAppointment) {
       return null;
     }
     String dateTime = optionalInstantToString(toHumanDateTime(rpcAppointment.dateTime()));
+    var participants = List.of(participants(), location(rpcAppointment.facility()));
```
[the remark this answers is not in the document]
bschofield-va: ^ this... keep decision making together and maintain the same level of abstraction with in a method.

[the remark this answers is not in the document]
jblefkove-va: Yes... it is also odd to have a plural `participants()` method returning a single participant

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
jblefkove-va: We assume `accepted` status for both patient and location participants in data-query, so I think it's reasonable to do the same here

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -82,12 +83,23 @@ String idFrom(String id) {
     return toResourceId(patientIcn, site, VprGetPatientData.Domains.appointments, id);
   }
 
-  List<Appointment.Participant> participants() {
+  Appointment.Participant location(CodeAndNameXmlAttribute facility) {
+    if (facility == null || facility.code() == null) {
+      return null;
+    }
+    var facilityIen = facility.code();
+    var display = "vha_" + site + "_" + facilityIen;
```
jblefkove-va, line 91: `facilityIen` is the VistA site ID, not the location IEN. Take a look at [this example](https://github.com/department-of-veterans-affairs/lighthouse-charon/blob/main/charon-models/src/test/resources/SampleAppointmentsResult.xml); the location IEN is `23`, which you have to parse out of `id.value`

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -154,7 +178,7 @@ private Appointment toAppointment(Appointments.Appointment rpcAppointment) {
         .serviceType(serviceType(rpcAppointment.clinicStop()))
         .appointmentType(appointmentType(rpcAppointment.type()))
         .start(dateTime)
-        .participant(participants())
+        .participant(participants(rpcAppointment))
```
jblefkove-va, line 181: if `id` is the only field used, why not pass that in here?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -82,12 +84,34 @@ String idFrom(String id) {
...
+        .build();
+  }
+
+  List<Appointment.Participant> participants(Appointments.Appointment rpcAppointment) {
+    return List.of(
+        Appointment.Participant.builder()
+            .actor(toReference("Patient", patientIcn, null))
+            .status(Appointment.ParticipationStatus.accepted)
+            .build(),
+        location(rpcAppointment.id()));
```
jblefkove-va, line 114: There needs to be protection here against `location()` returning null

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -82,12 +84,34 @@ String idFrom(String id) {
     return toResourceId(patientIcn, site, VprGetPatientData.Domains.appointments, id);
   }
 
-  List<Appointment.Participant> participants() {
+  Appointment.Participant location(ValueOnlyXmlAttribute id) {
+    if (id == null || id.value() == null || !id.value().contains(";")) {
+      return null;
+    }
+    var locationIen = id.value().substring(id.value().lastIndexOf(";") + 1);
```
jblefkove-va, line 91: probably also worth checking that `locationIen` is not blank

jblefkove-va (dismissed):
:shipit:

bschofield-va (changes requested):
blocking merge

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -82,12 +87,61 @@ String idFrom(String id) {
...
         .actor(toReference("Patient", patientIcn, null))
         .status(Appointment.ParticipationStatus.accepted)
-        .build()
-        .asList();
+        .build();
+  }
+
+  List<Appointment.Participant> participants(ValueOnlyXmlAttribute id) {
+    var participants = new ArrayList<Appointment.Participant>(2);
+    participants.add(participantPatient());
```
bschofield-va, line 142: @agarcia-va Note the change here where I normalize the level of abstraction in this method. It now delegates to two helpers instead of one helper and building one participant by hand.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -82,12 +87,61 @@ String idFrom(String id) {
...
+  Optional<Participant> participantLocation(ValueOnlyXmlAttribute id) {
+    var locationIen = locationIenFrom(id);
+    if (isBlank(locationIen)) {
+      return Optional.empty();
+    }
+    /*
+     * The Lighthouse Facilities API uses a prefixed ID. For healthcare facilities it's 'vha' plus
+     * the VAMC site plus the location (i.e. clinic).
+     */
+    var facilitiesApiIdentifier = "vha_" + site + "_" + locationIen;
```
bschofield-va, line 116: @agarcia-va note the explanation as _why_ we format the ID this way

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -82,12 +87,61 @@ String idFrom(String id) {
     return toResourceId(patientIcn, site, VprGetPatientData.Domains.appointments, id);
   }
 
-  List<Appointment.Participant> participants() {
-    return Appointment.Participant.builder()
+  String locationIenFrom(ValueOnlyXmlAttribute id) {
```
bschofield-va, line 90: @agarcia-va Note the extraction of location IEN parsing to it's own method. Also note the example in the comment so future stewards understand what kind of ID we were trying to parse. Finally note the handling of some edge case where the last character was a semicolon. The previous version would IndexOutOfBounds.

bschofield-va (approved):

PR #405, opened 2022-02-14
API-13384-effective-and-expiration-date

jhulbert-va:
# REQUIRES https://github.com/department-of-veterans-affairs/lighthouse-vistalink-mcode/pull/89


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/InsuranceBufferToR4CoverageTransformer.java
```diff
@@ -126,6 +135,20 @@ private Reference beneficiary(@NonNull String patientIcn, Optional<String> maybe
         .asList();
   }
 
+  private Period period(FilemanEntry entry) {
+    Period period = Period.builder().build();
+    entry
+        .internal(InsuranceVerificationProcessor.EFFECTIVE_DATE, this::toFilemanDate)
```
jabando-va, line 141: I think that field number needs to be added to `MAPPED_VISTA_FIELDS`. Same for `EXPIRATION_DATE`.

jabando-va (approved):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/R4Transformers.java
```diff
@@ -151,6 +151,19 @@ public static boolean isBlank(Object value) {
     return value == null;
   }
 
+  /** Verfies that a strings length is between two values. */
+  public static boolean isStringCorrectLength(int min, int max, String value) {
```
bschofield-va, line 155: Still trying to put my finger on why I dislike this method... perhaps it's the name, `correct` implies a right/wrong without the context in which we are asking the question. Perhaps it's the generic `min` and `max` which do not help me understand if the range is inclusive or exclusive of these values. Perhaps it's because it could be rewritten as a one-liner, e.g.,
```
boolean isStringLengthInRangeInclusively(int inclusiveMin, int inclusiveMax, String maybeString) {
  return Range.between(inclusiveMin, inclusiveMax).contains(StringUtils.length(maybeString));
}
```

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/R4Transformers.java
```diff
@@ -201,22 +214,26 @@ public static BigDecimal toBigDecimal(String string) {
...
+
   /** Transform a FileMan date to a human date. */
   public static Optional<Instant> toHumanDateTime(ValueOnlyXmlAttribute filemanDateTime) {
-    FilemanDate result = FilemanDate.from(filemanDateTime, ZoneId.of("UTC"));
-    if (isBlank(result)) {
+    if (filemanDateTime == null) {
       return Optional.empty();
     }
-    return Optional.of(result.instant());
+    return toHumanDateTime(filemanDateTime.value(), ZoneId.of("UTC"));
```
bschofield-va, line 236: we're defining the default zone of `UTC` in two different places, perhaps `toHumanDate(filemanDateTime.value())` is more appropriate?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/RequestPayloadExceptions.java
```diff
@@ -116,6 +116,20 @@ public String getMessage() {
     }
   }
 
+  public static class InvalidStringLength extends InvalidField {
+    @Builder
+    InvalidStringLength(String jsonPath, int minimum, int maximum, int received) {
+      super(
+          jsonPath,
+          "Value must be between "
```
bschofield-va, line 124: inclusively or exclusively?

eclendenning-va (approved):

bschofield-va (changes requested):
blocking merge

bschofield-va (approved):

PR #420, opened 2022-02-23
Add MedicationRequest reference to MedicationDispense

kivarsson-va:
Co-authored-by: Kira Ivarsson <kivarsson@rise8.us>
Co-authored-by: Michael Silverman <msilverman@rise8.us>


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -73,8 +75,16 @@
     return fill -> toHumanDateTime(fill.fillDate()).map(range::isDateWithinBounds).orElse(false);
   }
 
-  SimpleQuantity daysSupply(String fillDaysSupply) {
+  List<Reference> authorizingPrescription(String id) {
+    if (isBlank(id)) {
+      return null;
+    }
+    return List.of(
+        Objects.requireNonNull(
```
bschofield-va, line 83: We don't want NPE popping out because we failed to create a reference. You need to verify you are able to create the reference, returning `null` if not possible.

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformerTest.java
```diff
@@ -147,6 +148,12 @@ static R4MedicationDispenseTransformer _tx() {
     return Stream.of(Arguments.of(ignoredBecauseEmpty), Arguments.of(ignoredBecauseNoFillDate));
   }
 
+  @Test
+  void authorizingPrescription() {
```
bschofield-va, line 152: inadequate tests, please follow the pattern set by the other transformation methods that verify behavior of both happy and sad cases

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformers.java
```diff
@@ -27,6 +28,14 @@ public static Dosage dosageInstruction(String sig, String ptInstructions) {
         .build();
   }
 
+  /** Create a MedicationRequest id from and id, patientIcn and site. */
+  public static String medRequestIdFrom(String id, String patientIcn, String site) {
```
bschofield-va, line 32: needs tests

bschofield-va, line 32: Let's also clarify intention by being a little more verbose `medicationRequestIdFrom(String vistaMedId, ...`

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/MedicationDispenseSamples.java
```diff
@@ -91,6 +92,10 @@ public MedicationDispense medicationDispense(String id) {
                   .patientInstruction("take with food")
                   .build()
                   .asList())
+          .authorizingPrescription(
+              List.of(
+                  Objects.requireNonNull(
```
bschofield-va, line 97: why the `requireNonNull` here... you _know_ it's not `null` because it's hard coded.

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseWitnessProtectionAgentTest.java
```diff
@@ -23,6 +26,10 @@ void referencesOfT() {
             .id("md1")
             .meta(Meta.builder().source("123").build())
             .subject(Reference.builder().reference("Patient/p1").build())
+            .authorizingPrescription(
+                List.of(
+                    Objects.requireNonNull(
```
bschofield-va, line 31: See previous `requireNonNull` and `List.of` comments

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/MedicationDispenseSamples.java
kivarsson-va: yup, the `requireNonNull` was to appease the IDE, we can suppress the warnings if y'all are fine with them too

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformers.java
bschofield-va: `id` is too generic from a self-documenting-code perspective

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformersTest.java
```diff
@@ -94,6 +96,13 @@ void medicationCodeableConcept(Product product, CodeableConcept expected) {
     assertThat(actual).isEqualTo(expected);
   }
 
+  @Test
+  void medicationRequestIdFrom() {
+    assertThat(_tx.medicationRequestIdFrom("1234", "100053443", "673"))
+        .isEqualTo(R4Transformers.toResourceId("100053443", "673", Domains.meds, "1234"));
```
bschofield-va, line 102: This seems a bit weak to use a transformer to test a transformer. A bug in `toResourceId` which is used now in both the _actual_ and _expected_ would go undetected. Be explicit with the expected value.

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformerTest.java
```diff
@@ -147,6 +148,13 @@ static R4MedicationDispenseTransformer _tx() {
...
+    assertThat(_tx().authorizingPrescription(MedicationSamples.Vista.create().med().id().value()))
+        .isEqualTo(List.of(toReference("MedicationRequest", "sNp1+673+M33714", null)));
+    assertThat(_tx().authorizingPrescription("")).isEqualTo(null);
```
bschofield-va, line 155: `null` case

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformersTest.java
```diff
@@ -94,6 +96,13 @@ void medicationCodeableConcept(Product product, CodeableConcept expected) {
...
+    assertThat(_tx.medicationRequestIdFrom("", "111", "673")).isEqualTo(null);
```
bschofield-va, line 103: `null` cases

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformersTest.java
treynolds-va: All our method is doing is a null check before calling the other transformer so that seemed like the best way to test it. If the R4Transformer method is properly tested then this test shouldn't have any problems. If you would prefer a hardcoded _expected_ value then we can do that.

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformerTest.java
treynolds-va: Testing empty string and null seems like excessive testing of the _isBlank_ method that should already be tested elsewhere

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformersTest.java
treynolds-va: Same note as previous

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformerTest.java
bschofield-va: I prefer the file to appear as if one person authored it, i.e. I'd prefer the test to be parameterized like the other tests. I'll accept the mismatch in style as long as the test is covering cases the supporting method is likely to encounter, which is currently missing a null case. I believe the test should be ignorant of the implementation (usage of isBlank(...)) and focus on inputs/outputs, that way when someone decides to change the implementation (e.g. remove isBlank(...)) we don't get a surprise bug. The purpose of the test is verify the current implementation is behaving as expected _and_ to provide a safety net against future changes.

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformersTest.java
bschofield-va: please add the explicit hard coded expected value for same reason described in the comment below

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformersTest.java
bschofield-va: same response as previous

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformerTest.java
```diff
@@ -147,6 +148,13 @@ static R4MedicationDispenseTransformer _tx() {
     return Stream.of(Arguments.of(ignoredBecauseEmpty), Arguments.of(ignoredBecauseNoFillDate));
   }
 
+  @Test
+  void authorizingPrescription() {
+    assertThat(_tx().authorizingPrescription(MedicationSamples.Vista.create().med().id().value()))
+        .isEqualTo(List.of(toReference("MedicationRequest", "sNp1+673+M33714", null)));
```
bschofield-va, line 154: This suffers from the same problem as `toResourceId(...)` in the `R4MedicationTransformersTest`, Using `isEqualTo(Reference.builder().reference("MedicationRequest/sNp1+673+M33714").build().asList())` removes the circular "using production code to test production code" thing which can hide bugs, i.e. if I go and change `toReference(...)` to something stupid.

bschofield-va (approved):

bschofield-va:
❤️ thanks!

PR #431, opened 2022-02-28
API-000-production-to-v1

jhulbert-va:

Production does not yet have fall-back capability, so we do not want to move forward with deploying the conformance changes to that environment just yet:
https://libertyits.slack.com/archives/C02AQHDR0CQ/p1645802253208959


bschofield-va (approved):

jhulbert-va:
No one uses the openapi directly from VFQ

PR #436, opened 2022-03-04
cleanup fields to bring the read/write diff more in to line

eclendenning-va:


vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformerTest.java
```diff
@@ -355,6 +358,14 @@ void nameOfInsured() {
                         List.of(HumanName.builder().build(), HumanName.builder().build())));
   }
 
+  @Test
+  void organizationStatus() {
```
jhulbert-va, line 362:
```
@ParameterizedTest
@NullSource
@ValueSource(booleans = {false})

PR #446, opened 2022-03-10
Api 14191 buffer write add service date optional extension

ilaflamme-va:
- Tested with the new mcode PR

### DIFF
--- a/vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/ExtensionFactory.java
+++ b/vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/ExtensionFactory.java
@@ -9,10 +9,16 @@ import gov.va.api.health.r4.api.datatypes.Quantity;
 import gov.va.api.health.r4.api.elements.Extension;
 import gov.va.api.health.r4.api.elements.Reference;
 import gov.va.api.health.vistafhirquery.service.controller.definitions.MappableCodeableConceptDefinition;
+import gov.va.api.health.vistafhirquery.service.controller.definitions.MappableDateDefinition;
 import gov.va.api.health.vistafhirquery.service.controller.definitions.MappableExtensionDefinition;
+import gov.va.api.lighthouse.charon.models.FilemanDate;
 import gov.va.api.lighthouse.charon.models.lhslighthouserpcgateway.LhsLighthouseRpcGatewayResponse;
 import java.math.BigDecimal;
+import java.sql.Date;
+import java.text.SimpleDateFormat;
+import java.time.ZoneId;
 import java.util.Map;
+import java.util.function.Function;
 import lombok.AllArgsConstructor;
 
 @AllArgsConstructor(staticName = "of")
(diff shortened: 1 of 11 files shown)


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/ExtensionFactory.java
```diff
@@ -32,6 +38,12 @@ private Extension buildExtensionWithCodeableConcept(String value, String system,
         .build();
   }
 
+  private Extension buildExtensionWithValueDate(FilemanDate fm, String url) {
```
jhulbert-va, line 41: Perhaps this should be an Instant? Instant is a lot more robust than requiring us to always use a fileman date here?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -767,14 +778,28 @@ WriteableFilemanValue relatedPersonBirthDate(String birthDate) {
...
 
-  WriteableFilemanValue serviceDate() {
-    return factoryRegistry()
-        .get(InsuranceVerificationProcessor.FILE_NUMBER)
-        .forString(
-            InsuranceVerificationProcessor.SERVICE_DATE,
-            indexRegistry().get(InsuranceVerificationProcessor.FILE_NUMBER),
-            today())
-        .get();
+  WriteableFilemanValue serviceDate(List<Extension> extensions) {
```
jhulbert-va, line 781: Not really a fan of this. It basically has intimate knowledge that there is only one available extension. If we were to add another, it would required an entire refactor of the method. I think this should return a Collection of WFV's and make the decision whether to process or add service date as part of that collection.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/definitions/MappableDateDefinition.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+@Value
+@Builder
+public class MappableDateDefinition {
+  @NonNull String vistaField;
+  boolean isRequired;
+  @NonNull DateTimeFormatter dateFormatter;
```
jhulbert-va, line 13: Should this be `vistaDateFormatter`? Or something like that so there's no confusion as to what the format is representative of?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/extensionprocessing/DateExtensionHandler.java
```diff
@@ -0,0 +1,60 @@
...
+        filemanFactory,
+        definition.valueDefinition().vistaField(),
+        index);
+    this.dateFormatter = definition.valueDefinition().dateFormatter();
+  }
+
+  @Override
+  public List<WriteableFilemanValue> handle(String jsonPath, Extension extension) {
+    if (isBlank(extension.valueDate())) {
+      throw RequestPayloadExceptions.ExtensionMissingRequiredField.builder()
```
jhulbert-va, line 41: Import `ExtensionMissingRequiredField`

bschofield-va (approved):

PR #453, opened 2022-03-18
API-14284: update mapping spreadsheet part 2

jabando-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/R4CoverageToInsuranceBufferTransformer.java
```diff
@@ -528,12 +528,15 @@ WriteableFilemanValue patientId(Reference beneficiary) {
...
+                coding ->
+                    "MB".equals(coding.code())
+                        && "http://terminology.hl7.org/CodeSystem/v2-0203".equals(coding.system()));
     if (!isMemberId) {
       throw UnexpectedNumberOfValues.builder()
           .jsonPath(".beneficiary.identifier")
-          .identifyingFieldJsonPath(".type.coding[].code")
-          .identifyingFieldValue("MB")
+          .identifyingFieldJsonPath(".type.coding[].code && .type.coding[].system")
+          .identifyingFieldValue("MB && http://terminology.hl7.org/CodeSystem/v2-0203")
```
eclendenning-va, line 539: Remember where these values are used.

`message += format(" where %s matched %s.", identifyingFieldJsonPath, identifyingFieldValue);`

Maybe make this a little easier to read.

eclendenning-va (approved):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/InsuranceBufferDefinitions.java
```diff
@@ -96,7 +96,7 @@ public MappableIdentifierDefinition processorControlNumber() {
             MappableDateDefinition.builder()
                 .vistaField(InsuranceVerificationProcessor.SERVICE_DATE)
                 .vistaDateFormatter(DateTimeFormatter.ofPattern("MMddyyyy"))
-                .isRequired(false)
+                .isRequired(true)
```
jhulbert-va, line 99: While this field is _technically_ required, we can't mark it like that here or the extension processor will throw before we have a change to set a default value.

jhulbert-va (changes requested):
Blocking this merge for the one comment

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/coverage/InsuranceBufferDefinitions.java
jhulbert-va: We either need to refactor the transformer (which may be the move), or we need to keep this `false` in the definition.

PR #456, opened 2022-03-24
changed med dispense search param from date to whenprepared

msilverman-va:
FHIR standard APIs use resource field names as query params, so to reduce ambiguity in the search endpoint we renamed the `date` param to `whenprepared`. 

Co-authored-by: Kira Ivarsson <kivarsson@rise8.us>


bschofield-va (approved):

bschofield-va:
10/10 commit comment

PR #458, opened 2022-04-01
Add dosage extension for non-decimal dose strings

kivarsson-va:
Added/updated some tests and add a couple blank checks to the work Bryan did with this.

After looking at the CDW query results from KBS, we opted to leave out all parsing from the API since any parsing we need for our app is going to be pretty implementation-specific and probably isn't best for a broader API. Everything will just go into the originalText extension.


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformers.java
```diff
@@ -27,33 +28,46 @@
 
   /** Create a dosage from a sig and ptInstruction. */
   public static Dosage dosageInstruction(String sig, String ptInstructions, List<Dose> doses) {
+    List<DoseAndRate> doseAndRates = null;
 
-    if (allBlank(sig, ptInstructions, doses)) {
-      return null;
+    if (!isBlank(doses)) {
```
jhulbert-va, line 33: I think this is a bit over-complicated. `Safe.stream()` will create an empty stream if `doses` is `null`, ultimately ending with creating an empty list. The `allBlank` method on line 42 will check for null, as well as empty lists. 

Instead, you could set doseAndRates using Safe.stream, and use `emptyToNull(doseAndRates)` on line 49 to trim the list to null if it's empty.

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformersTest.java
```diff
@@ -20,14 +22,43 @@
...
 class R4MedicationTransformersTest {
-
   private final R4MedicationTransformers _tx =
       Mockito.mock(R4MedicationTransformers.class, Mockito.CALLS_REAL_METHODS);
 
   static CodeableConcept _cc(String text, List<Coding> codings) {
     return CodeableConcept.builder().text(text).coding(codings).build();
   }
 
+  static Dosage _dosage(
```
jhulbert-va, line 32: I think this should be two methods. It's creating a sample, so I don't think it should do complex logic. What about `_dosageWithExtension` and `_dosageWithDoseAndRate`?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformers.java
kivarsson-va: okay yeah thanks, we kept refactoring this to try to make it as uncomplicated as possible and without unnecessary data structures but i like your suggestion as a very straightforward way. forgot about the `emptyToNull` helper, too 🤦‍♀️

vista-fhir-query/src/test/java/gov/va/api/health/vistafhirquery/service/controller/medication/R4MedicationTransformersTest.java
kivarsson-va: agreed, will do

jhulbert-va (approved):

PR #459, opened 2022-04-04
Add extension for VistA medication status

kivarsson-va:
per KBS' status extension definition


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationrequest/R4MedicationRequestTransformer.java
```diff
@@ -154,4 +188,25 @@ private MedicationRequest toMedicationRequest(Med rpcMed) {
                     rpcMed.dose())))
         .build();
   }
+
+  Extension vistaStatus(String vistaStatus) {
```
jhulbert-va, line 192: Is there a reason we couldn't populate the display if the code doesn't map?

kivarsson-va: In Jay's coding definition, he was adhering to the VistA status codes specifically and didn't include an "unknown" option. We didn't specifically ask about unmapped values, but since he built out the status definition to explicitly be VistA-defined values, we were thinking we should only return values defined in the VistA spec.

image

jhulbert-va (approved):

PR #463, opened 2022-04-12
API-14952: medication date filter optimization

jabando-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationrequest/R4SiteMedicationRequestController.java
```diff
@@ -83,20 +89,31 @@ public MedicationRequest.Bundle medicationRequestSearch(
...
               value = "_count",
               required = false,
               defaultValue = "${vista-fhir-query.default-page-size}")
           int count) {
+    var fillFilter =
+        DateSearchBoundaries.optionallyOf(date)
+            .map(R4MedicationRequestTransformer::acceptOnlyWithDateInRange)
+            .orElse(acceptAll());
+    if (date == null || date.length == 0) {
+      date = new String[] {"ge1900"};
```
jhulbert-va, line 103: Same as the above comment. Are we sure that 1900 will not result in a timeout?

PR #478, opened 2022-05-16
Api 000 add date query to medication request

msilverman-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationrequest/R4MedicationRequestTransformer.java
```diff
@@ -98,12 +99,23 @@ public class R4MedicationRequestTransformer extends R4MedicationTransformers {
           .to("16")
           .build();
 
+  @NonNull final Predicate<Med> requestFilter;
```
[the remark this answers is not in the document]
msilverman-va: We form the predicate in the controller and pass it through to allow for other filter options in the future. It will be easier to add other conditions if we form the predicate in the controller rather than pass a lot of data through to the transformer. We do this for Medication Dispense too.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -193,7 +193,7 @@ public Stream<MedicationDispense> toFhir() {
                 Safe.stream(med.fill())
                     .filter(Objects::nonNull)
                     .filter(this::isViable)
-                    .filter(fillFilter())
+                    .filter(fillFilter)
```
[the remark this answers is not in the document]
kivarsson-va: there's no need to call the function when passing a predicate into a filter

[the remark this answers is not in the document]
jhulbert-va: We sometimes create lazy getters for fields. In case we were to do something like that for this field in the future, I would prefer it remain a method call.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationrequest/R4MedicationRequestTransformer.java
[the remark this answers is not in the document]
jhulbert-va: Right, I do see that's how it got implemented in MedicationDispense, however, I'm not sure that's correct. 

The controller requests data and bundles it, passing it off to the transformer to make decisions on the results. By building the predicate in the controller, the boundaries between those duties becomes less clear, the controller becomes aware of how the results will be filtered. 

By having the controller pass the DateSearchBoundaries to the transformer, those lines become clear again. The transformer is making the decision on what to do with the results based on the DateBoundaries is was passed by the controller.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -133,6 +138,18 @@ Optional<Integer> determineRemainingFills(
     return Optional.empty();
   }
 
+  private Predicate<Fill> dispenseFilter() {
+    if (isBlank(dateSearchBoundaries())) {
+      return fill -> true;
+    }
+    if (NumberUtils.isCreatable(dateSearchBoundaries()[0])) {
+      return acceptOnlyWithFillDateEqualTo(dateSearchBoundaries()[0]);
+    }
```
jhulbert-va, line 147: This bit seems odd to me. `DateSearchBoundaries` assumes `EQ` if no prefix is specified. This looks like it's doing the same.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -89,6 +87,13 @@ List<Reference> authorizingPrescription(String vistaMedId) {
     return isBlank(reference) ? null : reference.asList();
   }
 
+  /** Return the dateSeachBoundary if it exists. */
+  public String[] dateSearchBoundaries() {
```
jhulbert-va, line 91: Not sure you need this method. You do an `isBlank` check below.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -133,6 +138,18 @@ Optional<Integer> determineRemainingFills(
     return Optional.empty();
   }
 
+  private Predicate<Fill> dispenseFilter() {
+    if (isBlank(dateSearchBoundaries())) {
```
jhulbert-va, line 142: Optionally of will perform this check.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationrequest/R4MedicationRequestTransformer.java
```diff
@@ -124,6 +132,13 @@ List<CodeableConcept> category(String vaType) {
         .orElse(null);
   }
 
+  /** Return the dateSeachBoundary if it exists. */
+  public String[] dateSearchBoundaries() {
```
jhulbert-va, line 136: I think the same applies here.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
msilverman-va: The reason we have this is because on MedicationDispense read we have a FilemanDate i.e. "3110507" but on search we have a human readable date i.e. "2022-01-01". The acceptOnlyWithFillEqualTo allows us to check with the filemanDate rather than converting to human readable and then to DateSearchBoundaries.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
msilverman-va: I added the null check here because Arrays.copy requires dateSearchBoundaries.length and since dateSearchBoundaries could be null we get a null pointer exception when trying to grab it. We could remove this and change the controller to pass in an empty array when there are no search params.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
bschofield-va: Sorry, this is not going to work for me ... have a `String[] dateSearchBoundaries` that sometimes means a FHIR date and sometimes mean Fileman date is too easily broken. I think my team will take a crack at putting the finishing touches on this PR.

treynolds-va: If potentially having different string date types in dateSearchBoundaries is an unacceptable path then would it be acceptable for the medicationDispenseRead to convert the FilemanDate that it receives to a FHIR date before passing to the transformer?

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/DateSearchBoundaries.java
```diff
@@ -228,8 +228,7 @@ public boolean isDateWithinBounds(Instant date) {
     if (stop() == null) {
       return start().isBefore(date);
     }
-    return (start().isBefore(date) || start().equals(date))
-        && (stop().isAfter(date) || stop().equals(date));
+    return start().isBefore(date) && stop().isAfter(date);
```
kivarsson-va, line 231: Is there a reason this was changed back? We admittedly could've included this test to demonstrate why we did this, but right now these bounds aren't inclusive, so if you get someone querying a very specific date they're not going to get the expected results

image

kivarsson-va: that's fair. we had included it in this PR because we had a failing test without it, but it seems the test itself never actually made it into the PR.

we aren't too concerned about this change happening immediately, as we give our queries enough of a date buffer as it is that if we dropped something on the date boundary, it wouldn't be a huge deal. that said, we were looking around the codebase and it looks like it's only otherwise used for Conditions. based solely on what we've deduced from that logic, we made the assumption that the onset date for Conditions would also prefer an inclusive boundary. as you mentioned, though, it would be good to be sure it wouldn't negatively impact anything else

eclendenning-va (approved):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4MedicationDispenseTransformer.java
```diff
@@ -58,25 +58,19 @@ public class R4MedicationDispenseTransformer {
...
+  @NonNull String site;
 
-  @NonNull final String patientIcn;
+  @NonNull String patientIcn;
 
-  @NonNull final Predicate<Fill> fillFilter;
+  @NonNull Results rpcResults;
 
-  @NonNull final VprGetPatientData.Response.Results rpcResults;
+  Optional<DateSearchBoundaries> dateSearchBoundaries;
```
bschofield-va, line 67: Date search boundaries is ignored if fill data is specified. Ideally, we'd ensure they are mutually exclusive at construction, e.g. two different `@Builder` annotated private constructors.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/DateSearchBoundaries.java
bschofield-va: Hey @kivarsson-va, unless there is some business need for additional changes, I'd like to hold off on further changes. If a need exists, then let's get together and discuss then so my team can properly assess impact to existing behavior and future work before any additional code changes are made.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationdispense/R4SiteMedicationDispenseController.java
```diff
@@ -140,15 +134,19 @@ private MedicationDispenseId parseOrDie(String publicId) {
   }
 
   private R4Transformation<Results, MedicationDispense> transformation(
-      String site, String patientId, Predicate<Fill> fillFilter) {
+      String site,
+      String patientId,
+      Optional<DateSearchBoundaries> dateSearchBoundaries,
```
bschofield-va, line 139: I dislike this signature ...

jhulbert-va: Sounds good. I'll change it.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/DateSearchBoundaries.java
kivarsson-va: @bschofield-va of course, we're happy to chat about changes. we're trying to be as unintrusive as possible as we know you have a lot on your plates, but we are _always_ happy to chat and/or pair on this stuff.

and yes, we're good leaving this update off the PR. as for "business need," though, if the API supports an equivalence date check and it doesn't actually function properly, it seems like an issue regardless of the business need for our app specifically

bschofield-va (approved):

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/DateSearchBoundaries.java
bschofield-va: My team has recorded a defected in our backlog and will prioritize accordingly. Feel free to add addition information: 
https://vajira.max.gov/browse/API-16507

kivarsson-va: nope, looks straightforward to me. thanks a bunch!

PR #486, opened 2022-07-14
API-000-vfq-client-key

jhulbert-va:


vista-fhir-query-tests/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>gov.va.api.health</groupId>
     <artifactId>test-starter</artifactId>
-    <version>11.0.4</version>
+    <version>11.0.7-SNAPSHOT</version>
```
bschofield-va, line 7: whoa

bschofield-va (dismissed):

bschofield-va (approved):

PR #490, opened 2022-08-15
MedicationRequest pending -> draft status

ilaflamme-va:


ilaflamme-va:
@kivarsson-va

kivarsson-va:
do we want to remove [the other pending logic](https://github.com/department-of-veterans-affairs/health-apis-vista-fhir-query/blob/02b20f0cdc1b22a2198266320f1d6cb148a37b14/vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationrequest/R4MedicationRequestTransformer.java#L149)? or leave it in just in case something comes in with those medIDs?

bschofield-va:
> do we want to remove the other pending logic?

No, if we _know_ the record is from a Pending Outpatient Orders or Non-Verified Orders files, then we always want the status to be `draft`. If we can't determine the source file with certainty, then we'll evaluate the Vista provided status.

bschofield-va (approved):

PR #501, opened 2022-11-01
Api 206620 medication request identifier

ilaflamme-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/medicationrequest/R4MedicationRequestTransformer.java
```diff
@@ -209,6 +213,13 @@ Status fhirStatus(String vistaStatus, String vistaMedId) {
     return VISTA_STATUS_TO_STATUS.translate(vistaStatus).orElse(Status.unknown);
   }
 
+  private List<Identifier> identifiers(Identifier... identifier) {
+    if (identifier == null) {
+      return emptyList();
```
jhulbert-va, line 218: Should return null. We don't want to end up with an empty list.

jhulbert-va (approved):

PR #506, opened 2022-11-29
API-21315: MedicationRequest resource cleanup

jabando-va:


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/VistaSiteOidLookup.java
```diff
@@ -14,11 +14,12 @@
 
 /** Class for managing the mappings for oid and site. */
 @Component
-public class OidSiteMappings {
+public class VistaSiteOidLookup {
 
   private final Map<String, String> mappings;
```
jhulbert-va, line 19: Can we give this var a better name like `sta3nToOid`.

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/VistaSiteOidLookup.java
```diff
@@ -1,44 +1,39 @@
...
   }
 
   /** Find the matching site for a given oid. */
-  public String findSite(String oid) {
-    return mappings.get(oid);
+  public Optional<String> findSite(String oid) {
+    var inverse =
+        sta3nToOid.entrySet().stream()
+            .collect(Collectors.toMap(Map.Entry::getValue, Map.Entry::getKey));
+    return Optional.ofNullable(inverse.get(oid));
```
jhulbert-va, line 36:
```    
return 
        sta3nToOid.entrySet().stream()
            .filter(e -> oid.equals(e.getValue()))
            .findFirst();
```

vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/VistaSiteOidLookup.java
```diff
@@ -56,9 +51,10 @@ private Map<String, String> loadCsv(String mappingsCsv) {
     if (!file.exists()) {
       throw new IllegalStateException("Failed to load mappings file: " + mappingsCsv);
     }
-    // CSVRecord indexes: oid = 4, site = 1
+    // CSVRecord indexes: STA3N = 1, STATION_NUMBER (alphanumeric) = 2, OID = 4
     return CSVParser.parse(file, StandardCharsets.UTF_8, format).getRecords().stream()
-        .collect(Collectors.toMap(record -> record.get(4), record -> record.get(1)));
+        .filter(record -> record.get(1).equals(record.get(2)))
```
jhulbert-va, line 56: You don't end up with null pointers here at all? CSVParser might be returning empty strings, but we'll want to be careful here.

jhulbert-va (approved):

PR #510, opened 2023-01-26
API-000-serve-docs-from-root

jhulbert-va:
https://vajira.max.gov/browse/API-000


vista-fhir-query/src/main/java/gov/va/api/health/vistafhirquery/service/controller/metadata/R4MetadataController.java
```diff
@@ -41,7 +41,7 @@
 
 @RestController
 @RequestMapping(
-    value = {"/r4/metadata"},
+    value = {"/r4/metadata", "r4-metadata"},
```
eclendenning-va, line 44: `/r4-metadata`

eclendenning-va (approved):

